Issue #377, opened 2015-02-13
new user last login


kevinzurek:
When adding a new user, the last login should be never if that user hasn't logged in. Right now it's the account creation date.

kevinzurek:
Fixed in pr #379

PR #379, opened 2015-02-22
Changed last login to show "Never" when the user has never logged in.

emichaux:
I added an evolution to remove the not null constraint. The createUser method sets it to null. When creating the user item, If the last login is null, then it will set it to "Never".


kevinzurek:
Haven't gotten to testing it yet. Do those evolutions work on a database that isn't named 'femr'?

kevinzurek:
Confirmed they only work when the database is named femr, the downs throw an error when NULL age fields exist because it tries to make them NOT NULL

emichaux:
Yeah, I definitely overlooked that. Let me know what you think of this change.

kevinzurek:
Issue #377

PR #399, opened 2015-04-06
Feature conversion

aserhan:
Conversion Option to Metric


kevinzurek:
@aserhan if we move the metric setting check from the History, Medical, and Pharmacies controller to the retrieveVitalMultiMap function in the service layer then the conversion will happen everytime you need the vital map

aserhan:
So you suggest I do it that way, rather than the way I have it done now ?

app/femr/ui/controllers/HistoryController.java
```diff
@@ -102,7 +102,13 @@ public Result indexEncounterGet(int encounterId) {
         IndexEncounterMedicalViewModel indexEncounterMedicalViewModel = new IndexEncounterMedicalViewModel();
         IndexEncounterPharmacyViewModel indexEncounterPharmacyViewModel = new IndexEncounterPharmacyViewModel();
 
+        /* Alaa Serhan */
+        ServiceResponse<SettingItem> response = searchService.retrieveSystemSettings();
```
kevinzurek: Do we need to retrieve the settings here if all the patient/vital info is being converted in the service layer?

app/femr/ui/controllers/TriageController.java
```diff
@@ -136,6 +138,16 @@ public Result indexPost(int id) {
...
 
+        //Alaa Serhan
+        //Get the settings
+        ServiceResponse<SettingItem> settingItemServiceResponse = searchService.retrieveSystemSettings();
+        if (settingItemServiceResponse.hasErrors()) {
+            throw new RuntimeException();
+        }
+
+
+        SettingItem settings = settingItemServiceResponse.getResponseObject();
```
kevinzurek: this never gets used!

app/femr/ui/views/medical/edit.scala.html
```diff
@@ -30,16 +30,25 @@
...
+            } else {
+                <input type="hidden" id="vitalUnits" value="imp" />
+            }
+            
             <div id="vitalsWrap">
+                @if(viewModel.getSettings.isMetric) { <!-- Alaa Serhan -->
+                    <input type="hidden" id="vitalsUnits" value="metric" />
+                } else {
+                    <input type="hidden" id="vitalUnits" value="imp" />
+                }
```
kevinzurek: What is the purpose for these hidden input fields?

app/femr/ui/views/medical/listVitals.scala.html
```diff
@@ -57,15 +63,44 @@
...
+                        vitalMap.get("heightInches", vitalMap.getDate(dateIndex - 1)),
+                        viewModel.getSettings.isMetric,
+                        ""
+                    )
+                }
+                </td>
+            <!---  Alaa Serhan Must Format Output -->
+@*          @for(dateIndex <- 1 to vitalMap.getDateListChronological.size) {
+                    <td>@vitalMap.get("heightFeet", vitalMap.getDate(dateIndex - 1))
+                        . @vitalMap.get("heightInches", vitalMap.getDate(dateIndex - 1))</td>*@
```
kevinzurek: Can this be removed?

app/femr/ui/controllers/TriageController.java
aserhan: I will fix that. It was being user previously. I left it behind when I moved everything to Service Layer.

app/femr/ui/views/medical/edit.scala.html
aserhan: I have to test this again and fix it. This may no longer be needed. They are referenced in js.

app/femr/ui/views/medical/listVitals.scala.html
aserhan: If you mean the commented out code, yes it can be removed. However, the non commented code is to format the display for metric or imperial.

app/femr/ui/controllers/HistoryController.java
aserhan: We still need the settings in the view to output the text and display proper units when it Is imperial or metric. If you have an idea on how we can do the formatting without it I'm ready to change it.

The only thing I can do is add a variable to models..

But that's just moving stuff around.

Not really a solution...

Let me know if you have any suggestions.

kevinzurek: I see! :)

aserhan:
Are there any other comments that you have for this? Other than removing some of the things left behind.

kevinzurek:
The only thing i can find right now is that recording a new weight on the medical page breaks the BMI score. Looking into this

aserhan:
I will make sure to work on this tonight! Thank You!

aserhan:
That is the same issue whether you ONLY enter weight or ONLY enter height.

The BMI doesnt get calculated.

If a patient was initially admitted into triage and both weight and height were entered and we adding new vitals in Medical we can get the last height or weight entered if one of them is missing, and use that to calculate the BMI.

However you have the same issue in Triage, when you create a NEW patient. You can only submit weight if you want to or only submit height. But what happens if there was never height entered or never a weight entered

kevinzurek:
Look at the entries on medical as vital replacement. If height and weight are entered on triage and weight is replaced on medical, we still have height and weight for the bmi calculation

kevinzurek:
manually merged into development

PR #402, opened 2015-04-15
Feature pdf

AndreFarah:
PDF Feature


kevinzurek:
Initially just a few things i can find -

Can we change the button name from "Print" to "PDF" or "Generate PDF" since we're not actually printing with the button?

Also the controller is named "PDFConversionDemo", but it's not a demo!

conf/routes and conf/applications.conf both have commits with either formatting or comments in them. 

Other than that, great job on the re-base and facilitating multiple chief complaints!

AndreFarah:
Hey Kevin,

thank you for the feedback.

i have fixed the issues , but i didn't find any comments in my routes files
that i should get rid off.

kevinzurek:
Not so much comments, but see the diff here:

https://github.com/FEMR/femr/pull/402/files#diff-93c1ebbcb555c5f7d8ee08519974efa6

AndreFarah:
they are a copy of the same exact thing, the only difference i found is
that the research routes were duplicated, i got rid of it , also, i changed
the PDF route just now, so instead of showing in the browser "/PDFtest/2/4"
(for example)
it wouls now show :" /PDF/Pid=2/Eid=14 "  where Pid is Patient id and Eid
is Encounter id

PR #403, opened 2015-04-15
Feature revise encounter

aserhan:
Hello, 

I Fetched Amney's CurrentEditEncounter from her Fork. 

I did the cherry picking for my commit. 

I did the steps to rebase development etc..

I fixed conflicts. 

I compiled. 

Fixed Errors.

Compiled again. 

Let me Know if the Pull Request is Correct Now .

Thank You! Alaa


kevinzurek:
If you create one chief complaint for a patient and close their encounter then edit any of the fields under the chief complaint. The edits aren't reflected when the page is reloaded.

Also odd behavior if you do this with multiple chief complaints existing for that encounter.

conf/evolutions/default/9.sql
```diff
@@ -1,12 +1,12 @@
...
+                                    PRIMARY KEY (`id`) ,
+                                    UNIQUE INDEX `id_UNIQUE` (`id` ASC) ,
+                                    UNIQUE INDEX `name_UNIQUE` (`name` ASC) );
 
-# --- !Downs
+                                  # --- !Downs
 
-DROP TABLE IF EXISTS `medications`
\ No newline at end of file
+                                  DROP TABLE IF EXISTS `medications`
```
kevinzurek: this is only formatting and old evolution files shouldn't be altered

aserhan:
That is the same issue whether you ONLY enter weight or ONLY enter height.

The BMI doesnt get calculated.

If a patient was initially admitted into triage and both weight and height were entered and  we  adding new vitals in Medical we can get the last height or weight entered if one of them is missing, and use that to calculate the BMI. 

However you have the same issue in Triage, when you create a NEW patient. You can only submit weight if you want to or only submit height. But what happens if there was never height entered or never a  weight entered  .

aserhan:
Hello,
We resolved the issues you mentioned, however could not think of a solution to do Diagnoses, Dispensed Medications and the Chief Complaint name the same. There is no "History" like there is for the tab fields. You add a prescription/diagnoses or delete it. 
Do these 3 fields under the Assessment also need to be editable? If so, do you have any advice/solutions that may help in the implementation?

kevinzurek:
I think that what is editable will suffice for now. Ideally we will be able to edit the chief complaint itself, but i don't think this is an easy task and can be done in the future.

When you have 2 chief complaints, try changing one field such as Onset. It seems to change under both chief complaints (until you refresh the page).

app/femr/ui/views/history/editEncounterPost.scala.html
```diff
@@ -0,0 +1,17 @@
...
+@additionalStyles = {
+    <link rel="stylesheet" href="@routes.Assets.at("css/history.css")">
+}
+
+
+@main("History", scripts = additionalScripts, styles = additionalStyles) {
+
+
+
+}
```
kevinzurek: is this file needed?

app/femr/ui/views/history/editEncounterGet.scala.html
```diff
@@ -0,0 +1,310 @@
...
+
+@*
+@inputText("First Name", "firstName", true, if(viewModel != null) viewModel.getPatientItem.getFirstName else null, "text")
+@inputText("Last Name", "lastName", true, if(viewModel != null) viewModel.getPatientItem.getLastName else null, "text")
+@inputText("Address", "address", false, if(viewModel != null) viewModel.getPatientItem.getAddress else null, "text")
+@inputText("City", "city", true, if(viewModel != null) viewModel.getPatientItem.getCity else null, "text")
+@inputText("Age", "Years", true, if(viewModel != null) viewModel.getPatientItem.getAge else null,"text")
+@inputText("Sex", "sex", true, if(viewModel != null) viewModel.getPatientItem.getSex else null,"text")
+
+    *@
```
kevinzurek: i can't find the use for this file!

app/femr/ui/controllers/HistoryController.java
```diff
@@ -8,15 +10,21 @@
...
 import femr.ui.views.html.history.indexPatient;
+import femr.ui.views.html.history.editEncounterGet;
+import femr.ui.views.html.history.editEncounterPost;
+import femr.ui.views.html.history.listTabFieldHistory;
 import femr.util.DataStructure.Mapping.TabFieldMultiMap;
 import femr.util.DataStructure.Mapping.VitalMultiMap;
+import femr.util.stringhelpers.StringUtils;
+import play.data.Form;
 import play.mvc.Controller;
+import play.mvc.Http;
```
kevinzurek: a lot of these imports are not used

app/femr/ui/controllers/HistoryController.java
```diff
@@ -29,24 +37,31 @@
...
+    private final Form<IndexPatientViewModelPost> EditEncounterForm = Form.form(IndexPatientViewModelPost.class);
+    private final Form<fieldValueViewModel> fieldValueViewModelForm = Form.form(fieldValueViewModel.class);
+    private final IMedicationService medicationService;
     private final IEncounterService encounterService;
     private final ISessionService sessionService;
     private final ISearchService searchService;
     private final ITabService tabService;
     private final IPhotoService photoService;
     private final IVitalService vitalService;
+    private final IEncounterService assessmentService;
```
kevinzurek: some of the added services are not used

aserhan:
Okay I fixed it again. It should work better now.

kevinzurek:
Two things stand out:

1)The encounter field doesn't get updated after you change it. It stays "N/A" until you refresh the page. I think this could confuse a user because it doesn't indicate something was saved.

2)The date of medical visit is updated when you change an encounter field. This is an issue for 2 reasons: you don't know the actual date/time that the patient was seen in medical. The second reason is a little more complicated - If a patient is seen in medical and is not seen in pharmacy, the encounter will close in 2 days time. If you have a patient like this and you edit their encounter, the encounter will be re-opened because of the updating of 'date_of_medical_visit'.

For thing 2, removing line 374 in the history controller might help (;. There's also some unused services, methods, and imports in this controller. If you are using default IntelliJ key bindings you can find them by pressing f2.

kevinzurek:
manually rebased and merged. Solid work, everyone!

PR #404, opened 2015-04-15
Feature inventory

aserhan:
This is the Pull Request for Inventory.

I did the steps below.

However, I had no conflicts. 

I did a couple time to make sure. 

I compiled - No errors

I forced pushed and that it. 

Anyways let me know if all is good. 

Thank You!

Steps
git checkout development
git pull upstream development
git checkout feature
git rebase development
git status
fix any conflicts
git add file(or git add .)
git rebase --continue
git status
etc until done


kevinzurek:
Try cherry picking commit 719a05bd4bf71359607fb3f119e44a43f651108d from the inventory branch on my fork. This should fix evolution 75 for you, let me know if it doesn't work.

Did you need to seed values for the medication administrations in the DatabaseSeeder?

aserhan:
Okay I will try to do that. And yes please. I'm going to need your help with the DatabaseSeeder

kevinzurek:
Use the diagnosis seeder on line 85 here for reference:

https://github.com/FEMR/femr/blob/master/app/femr/util/startup/DatabaseSeeder.java

kevinzurek:
Pulled down your last couple commits and awesome!!! going to test more tomorrow!

kevinzurek:
As discussed:

-2 medications with only different ingredients should both be distinguishable
-error on medical and pharmacy when submitting and prescription fields are blank
-changing from BIW then back to BID will not produce the right amount any longer
-history is not showing the comprehensive dispensed prescription

aserhan:
So for the first issue we should add ingredient ONLY to distiguish it ? Do you also want this information listed in the typeahead? Should ALL the ingredients be listed along with it?

kdunlap:
Without being able to test the code at the moment, I will try best to answer your questions.

1. Add whatever is needed to the medication name so that it can be distinguished from others. If the inventory logic allows two medications with the same ingredients, then you probably need something else so they can be distinguished. I think the amount/ratio of the ingredients is more important than the ingredients themselves.

2. This information should also be listed in Typeahead, or anywhere else they will need to be distinguished for that matter. A user will use typeahead to select the patients prescriptions and will need to know exactly what they are choosing. You can do custom styles to the typeahead suggestions -- see the patient search on the homepage.

3. See answer 1.

I hope this helps.

kevinzurek:
1) The amount calculation isn't working for me on medical (if you enter BID x3, amount =  0)
2) I don't think adding a new field for ingredients is necessary. If we append it to the name of the medication, this should let someone search for "hydroco----" and see all the different kinds of hydrocodone related medication available.
3) The pharmacy needs to see the instructions. If the amount is 10, we don't automatically know if it is BID or BIW or whatever else it might be.
4) The PDF feature was merged into development - if you rebase and view the PDF, the prescription is not reflected here.

I would recommend instead of splitting the prescription displaying field into amount and name, just have one string called "prescription" that displays the entire prescription with amount, frequency, name, ingredients. This will make it easier on pdf and pharmacy. It can be created when the item is created in the service layer and you'll always have it in the UI layer.

aserhan:
Number 2  in regards this comment what do u mean can you be more specific?

kevinzurek:
When i say "field for ingredient" im referring to the ingredient column in the table on the admin screen. The name of the medication can include the ingredients (in the UI). When searching on the typeahead in medical/pharmacy, i think this would allow the user to search for the ingredient and also get the name. does that make more sense?

aserhan:
I don't understand what you mean for ingredients. 

Do you not want it seperate in database also? 

Even in the table in admin you said append to the name but what happens if there is multiple ingredients? 

Or what happens if there is the same ingredients with different  values for the measurements.

Do you want it all this appended?

kevinzurek:
Nope, i'm just referring to appending it in the service layer. The database wouldn't change. Here's a screenshot of what i'm talking about in the name column:
inventory

Perhaps make the ingredients a font size smaller? I think this would let you completely remove the ingredients column. You would also be able to display the medications this way in the typeahead and make the ingredient part of the searching.

aserhan:
All Changes have been made. 

In regards to the BID it's working on my end perfectly fine just make sure you have these values in the database table medication administrations under daily modifiers. 
daily modiefier

kevinzurek:
I have not looked, but they should be in the database seeder if they are necessary. Otherwise no one else will get them

aserhan:
I did add it to the Database Seeder.

aserhan:
Sorry I didn't mean to close

app/femr/util/startup/DatabaseSeeder.java
```diff
@@ -1042,6 +1044,149 @@ private void seedDefaultTabFieldSizes() {
...
+                medicationAdministration.setName("TID");
+                medicationAdministration.setDailyModifier(3f);
+                medicationAdministrationsToAdd.add(medicationAdministration);
+            }
+            if (!containMedicationAdministration(administrations, "TIW")) {
+                medicationAdministration = new MedicationAdministration();
+                medicationAdministration.setName("TIW");
+                medicationAdministration.setDailyModifier(0.42857f);
+                medicationAdministrationsToAdd.add(medicationAdministration);
+            }
```
kdunlap: The daily modifier values listed in your screenshot and those in the database seeder do not match up. Am I missing something?

aserhan:
Sorry for the confusion, this is what I had initially in the DB before the Seeder and some of the values are incorrect. You should have the values that are in the seeder.

kevinzurek:
The issue was because the updated seeder had the correct values and wasn't replacing the incorrect ones with the correct ones. The typeahead is also working!

After a rebase, the PDF is only displaying the medication names instead of the entire prescription right now. If we add a 'full name' field to prescription items just like a medicationitem has, then that can be used in both history and PDF - it would also eliminate view logic like we find on line 149 in history/indexEncounter.scala.html.

kdunlap:
A couple of quick things I found that are issues with existing prescription data.

Since the prescription were were a free-form field previous to this feature. We have to account for that data being in the database. The free form prescriptions will have a null in the medication_form column, so we need to add that check to all medication queries in the services.

Add .isNotNull("medicationForm") to all getMedicationQuery
- InventoryService around Lines: 87-89, 124-126, 315-318
- Lines 315-318 should also have eq(“isDeleted”, false) then the sorting by isDeleted can be removed
- Also need to add this to the Typeahead lists, so only inventoried prescriptions are suggested to the user
- MedicationService Lines: 282-284, 304-306

aserhan:
Okay I will get to these sometime this coming week .

kevinzurek:
Squashed all the work that has been done up to this point and rebased with development. 

Up-to-date branch:
https://github.com/kevinzurek/femr/tree/feature-FEMR25-medication_tracking

JIRA issue for medication tracking:
https://teamfemr.atlassian.net/browse/FEMR-25

PR #407, opened 2015-04-24
UpdateSex refactor

kdunlap:
Removed check that only let patient sex be updated if it was null previously. Added check to return error if patient does not exist.


kevinzurek:
I can't remember how we left off on this. Are we waiting for me to review and merge?

kdunlap:
I meant to merge this. I'll just go ahead and do it. We talked about the UI preventing the changing of a patient sex, but the Service layer should be able to allow an already defined patient sex to be changed. I just wanted to make sure before merging.

PR #408, opened 2015-05-03
creating non-generic repositories

kevinzurek:
@kdunlap I've opened this pull request prematurely so we can keep track of what's going on - i merged your implementation of the patient encounter repository here.

I changed the list element types that are being returned from repositories from List<? extends E> to List< E >. I think it might be cleaner to do the "casting" in the repositories


kevinzurek:
Currently the service layer will have no way of distinguishing between a response from the data layer as an error or as an entity not existing in the database....

kevinzurek:
rebased this with development

kevinzurek:
rebased again

kevinzurek:
closed because this is on hold

PR #419, opened 2015-06-21
Adding uglify, digest, and gzip for production builds.

emichaux:
This should compress the JS files, and create asset fingerprinting.
From the Play Documentation, "In essence asset fingerprinting permits your static assets to be served with aggressive caching instructions to a browser. This will result in an improved experience for your users given that subsequent visits to your site will result in less assets requiring to be downloaded."

This still needs some more testing to make sure no Javascript is broken after compressing.


kevinzurek:
It worked for me! 

I get a ton of warnings about what looks like not following best practices in javascript

kevinzurek:
I wasn't able to fix the warnings coming out of Uglify. I removed Uglify and merged this manually so we have Digest.

The migration guide from Play 2.3 also mentioned an UglifyJS 2 plugin soon.

PR #425, opened 2015-07-19
HTTP Filters

kevinzurek:
-Create an HTTP Filter to help control the research only feature. 
-Add a research only flag to application.conf.
-Update JDK and Scala version in Travis CI.


kdunlap:
I think this is great. It is clean and pretty future proof - meaning if we add new pages/features this will continue to work how we want it to.

PR #438, opened 2015-09-30
Users and trips (FEMR60 progress)

kevinzurek:
Restructure the settings menu, remove the concept of current trip from schema, and implement schema for tracking which users were on a trip.

Comment out mockdatamodelmapper


kdunlap:
Noting that this PR will temporarily break the home page user search. This feature will be fixed in a future PR. Just need to complete the User/Trip association feature.

kevinzurek:
image

PR #441, opened 2015-11-11
Weeks pregnant

zanni022:
[FEMR-23] and [FEMR-11]
-Added weeksPregnant to the vitals table
-Added weeksPregnant to the medical section


conf/evolutions/default/85.sql
```diff
@@ -0,0 +1,12 @@
+# --- !Ups
+
+INSERT INTO `vitals` (`id`, `name`, `data_type`, `unit_of_measurement`)
+  VALUES ('11', 'weeksPregnant', 'float', 'weeks');
```
kdunlap: Great job making this evolution! One small thing though - We don't really want to manually insert the id. MySQL will auto increment the Id for us. Remove the id portion of the insert and the rest should be good.

Even though assigning the Id of 11 should be safe, we still want this evolution to work if for some reason id 11 already exists in someone's femr install. Some early evolutions do not follow this though.

conf/evolutions/default/85.sql
```diff
@@ -0,0 +1,12 @@
...
+
+# --- !Downs
+
+DELETE FROM `patient_encounter_vitals`
+WHERE `vital_id` = 11;
```
kdunlap: After changing the above this should probably be:

DELETE FROM `patient_encounter_vitals` WHERE `vital_id` IN (SELECT id FROM vitals WHERE name = 'weeksPregnant')

conf/evolutions/default/85.sql
```diff
@@ -0,0 +1,12 @@
...
+
+DELETE FROM `vitals`
+WHERE `id` =11;
```
kdunlap: DELETE FROM `vitals` WHERE `name` = 'weeksPregnant`

conf/evolutions/default/85.sql
zanni022: Great! Thanks so much Ken!

app/femr/ui/controllers/TriageController.java
```diff
@@ -161,8 +161,11 @@ public Result indexPost(int id) {
 
 
         //create and save a new encounter
-        PatientEncounterItem patientEncounterItem =
-                populatePatientEncounterItem(viewModel.getChiefComplaint(), viewModel.getChiefComplaintsJSON(), viewModel.getWeeksPregnant(), currentUser, patientServiceResponse.getResponseObject().getId(), viewModel.getAgeClassification());
+        // PatientEncounterItem patientEncounterItem =
+                // populatePatientEncounterItem(viewModel.getChiefComplaint(), viewModel.getChiefComplaintsJSON(), currentUser, patientServiceResponse.getResponseObject().getId(), viewModel.getAgeClassification());
+        // ServiceResponse<PatientEncounterItem> patientEncounterServiceResponse = encounterService.createPatientEncounter(patientEncounterItem);
```
kdunlap: Its great to comment out the previous code while working on new stuff, but I think once the new code is working, go ahead and delete the previous stuff. If we need to see what was there before, we can go back in the git history :)

conf/evolutions/default/85.sql
kdunlap: also -- data_type can probably just be int. I'm not sure if anyone would need to record 9.5 weeks pregnant or not. Thoughts?

kdunlap:
Everything here looks great! If you make those small changes to your local branch and push them to your fork, the pull request should update automatically. Then we can merge this code into development

zanni022:
Hey Ken,

Thank you for all of the help! I just pushed them to my fork so they should
be there :) Let me know if it doesn't show up! Thanks so much!

Sincerely,
Samantha

PR #444, opened 2015-11-14
Improvement femr60 users and trips (incomplete)

kevinzurek:
-Lets the superuser add users to a trip
-Removes the default admin user from the user list
-Fixes broken patient searching from the last pull request i did

### DIFF
--- a/app/femr/business/services/core/IMissionTripService.java
+++ b/app/femr/business/services/core/IMissionTripService.java
@@ -27,6 +27,15 @@ import java.util.List;
 public interface IMissionTripService {
 
     /**
+     * Adds a list of users to a trip.
+     *
+     * @param tripId id of the trip to add users to, not null
+     * @param userIds list of user ids being added to the trip, not null
+     * @return the current trip or null if an error occurs
+     */
+    ServiceResponse<MissionTripItem> addUsersToTrip(int tripId, List<Integer> userIds);
+
+    /**
      * Retrieve the current trip information.
      *
      * @return the current trip or null if an error occurs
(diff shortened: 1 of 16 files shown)


kdunlap:
The SuperUser admin page for editing trips bugs out with the cities dictionary in the mission_cities table (90k entries). While this is an extreme case and we've already decided that the dictionary probably shouldn't live in the mission_cities table, its worth looking into for the long term.

kevinzurek:
Right now, the country is auto-populated based on the city that you choose (fEMR doesn't have more than 10 cities in it outside the dictionary). If we have the user choose the country first, the cities can be populated based on the country.

On another note, the superuser currently has the ability to add cities to the mission_cities table so that it is possible to make sure the city they need exists when linking up to a trip (/superuser/trips/cities). If we add the city dictionary to this table, could we eliminate this functionality completely since, at least in theory, all the cities anyone could ever use would exist? I am not saying that it still shouldn't exist separately in the cleanse-generator... just questioning if it should exist in the mission_cities table inside fEMR. In fact, i think this is why we discussed uploading the GPS coordinates of the cities as well - so we could programatically map out where fEMR is used. 

I'm trying to tag AJ in this conversation, but it won't let me!!

kdunlap:
This makes sense. Gonna merge this since its not really a problem until we put a list of cities with Lat,Lng coords in there.

AJSclyn:
We are currently still in the works on adding to the mission_cities_table when the user clicks Other in the Cleanse Generator.  The problem that we'll run into is that the user needs to know the specified country ID to enter as well.  I've provided an update to the sql queries to include latitude and longitude coordinates that can be run into a fresh table.
image

kevinzurek:
The current scope of your work for the cleanse-generator is to hook up to the mission_cities table inside of the fEMR database, correct? If this is the case, we should do a pull request to get the city data loaded in especially for testing the city typeahead on Triage and the city selection that was the original topic of this pull request.

PR #445, opened 2015-11-14
 Added 2 usertrackable columns for DateCreated and UserCreated

zanni022:
Created two trackable columns date_created and user_created to track who creates the new user and when they create the new user. FEMR TASK IN JIRA : [FEMR-51]

### DIFF
--- a/app/femr/data/models/mysql/User.java
+++ b/app/femr/data/models/mysql/User.java
@@ -55,6 +55,10 @@ public class User implements IUser {
     private Boolean passwordReset;
     @Column(name = "notes")
     private String notes;
+    @Column (name = "creation_date", nullable = false) //Sam Zanni
+    private DateTime DateCreated; //Sam Zanni
+    @Column (name = "user_created", nullable = false) //Sam Zanni
+    private String UserCreated; //Sam Zanni
     @ManyToMany(fetch = FetchType.LAZY, targetEntity = MissionTrip.class)
     @JoinTable(
             name = "mission_trip_users",
@@ -167,6 +171,18 @@ public class User implements IUser {
     public void setNotes(String notes) {
         this.notes = notes;
     }
+
+    @Override
+    public DateTime getDateCreated() { return DateCreated; }
+
+    @Override
+    public void setDateCreated(DateTime DateCreated) { this.DateCreated = DateCreated; }
+
+    @Override
+    public String getUserCreated() { return UserCreated; }
+
+    @Override
+    public void setUserCreated(String UserCreated) { this.UserCreated = UserCreated; }
 
     @Override
     public List<IMissionTrip> getMissionTrips() {
--- a/app/femr/data/models/mysql/User.java
+++ b/app/femr/data/models/mysql/User.java
@@ -55,10 +55,10 @@ public class User implements IUser {
     private Boolean passwordReset;
     @Column(name = "notes")
     private String notes;
-    @Column (name = "creation_date", nullable = false) //Sam Zanni
+    @Column (name = "date_created", nullable = false) //Sam Zanni
     private DateTime DateCreated; //Sam Zanni
-    @Column (name = "user_created", nullable = false) //Sam Zanni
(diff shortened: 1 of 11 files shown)


app/femr/data/models/mysql/User.java
```diff
@@ -55,6 +55,10 @@
     private Boolean passwordReset;
     @Column(name = "notes")
     private String notes;
+    @Column (name = "creation_date", nullable = false) //Sam Zanni
+    private DateTime DateCreated; //Sam Zanni
+    @Column (name = "user_created", nullable = false) //Sam Zanni
+    private String UserCreated; //Sam Zanni
```
kdunlap: I think that UserCreated should link to an actual user_id rather than a string of the first and last name. I also think UserCreated is kind of ambiguous, maybe change it to CreatedBy?

```
private User CreatedBy;
```

This change would affect the migrations as well. The user_created column in the migration would be something like:

```
ADD COLUMN `created_by` INT NOT NULL AFTER `creation_date`;
CONSTRAINT `fk_users_created_by_users_id`
    FOREIGN KEY (`user_id`)
    REFERENCES `users` (`id`));
```

Not sure if this is 100% correct. The downside is that we'd be adding a self-referencing column to the users table. I know we've had some weird issues with this in the past in regards to prescriptions and replacements. We might want to talk about this @kevinzurek: do you have any input here?

zanni022: Hey Ken!

Okay, thanks! I will start making those changes and I will also wait to
hear Kevin's input before I finalize the changes. Thanks again!

Sincerely,
Samantha

zanni022: Hey Ken,

I pushed some more changes for the trackable columns in the same way that I
did the new changes for the weeksPregnant. I tried doing the evolution
within the migration the way that you sent it over but had several errors
and would not work, therefore, I kept the old migration that I had and
switched the variable names. If you could, could you please look this over?
Thanks! If I can't figure out the new migration, would it be possible to go
over this on Tuesday at the meeting? Thank you and have a good rest of the
day!

Sincerely,
Samantha

conf/evolutions/default/85.sql
```diff
@@ -0,0 +1,14 @@
+# --- !Ups
+
+ALTER TABLE `users`
+ADD COLUMN `date_created` DATETIME NOT NULL AFTER `last_login`;
+ADD COLUMN `created_by` DATETIME NOT NULL AFTER `date_created`;
```
kdunlap: Change type of created_by to Int

app/femr/data/models/mysql/User.java
```diff
@@ -55,6 +55,10 @@
     private Boolean passwordReset;
     @Column(name = "notes")
     private String notes;
+    @Column (name = "date_created", nullable = false) //Sam Zanni
+    private DateTime DateCreated; //Sam Zanni
+    @Column (name = "created_by", unique = true, nullable = false) //Sam Zanni
+    private Integer CreatedBy; //Sam Zanni
```
kdunlap: Should be: 

```
private User CreatedBy;
```

This is a little trickier because we need to define the relationship so eBean can eagerly fetch the attached User if needed. We could talk about this on Tuesday

conf/evolutions/default/85.sql
```diff
@@ -0,0 +1,14 @@
+# --- !Ups
+
+ALTER TABLE `users`
+ADD COLUMN `date_created` DATETIME NULL AFTER `last_login`;
+ADD COLUMN `created_by` DATETIME NOT NULL AFTER `date_created`;
```
kdunlap: Change:

ADD COLUMN `created_by` DATETIME NOT NULL AFTER `date_created`;

To:

ADD COLUMN `created_by` INT NOT NULL AFTER `date_created`;

Then we can re-test, but looks good otherwise.

kdunlap:
Replaced by PR461 - closing.

PR #450, opened 2015-11-22
Edited evolution 85

kdunlap:
Changed evolution 85 to migrate existing weeks_pregnant data on both the ups and downs


kdunlap:
Need to remove weeks_pregnant from the PatientEncounter model. Got a little overzealous...

PR #451, opened 2015-11-23
Improvement FEMR92 edit evolution 85

kdunlap:
* Added data migration to Evolution 85 (moving weeks pregnant to vitals table)
* Removed weeksPregnant from the PatientEncounter model and items
* Fixed getting weeks pregnant within research.

### DIFF
--- a/app/femr/business/helpers/QueryHelper.java
+++ b/app/femr/business/helpers/QueryHelper.java
@@ -45,6 +45,21 @@ public class QueryHelper {
         return weight;
     }
 
+    public static Integer findWeeksPregnant(IRepository<IPatientEncounterVital> patientEncounterVitalRepository, int encounterId){
+        Integer weeks_pregnant = null;
+        Query<PatientEncounterVital> query2 = QueryProvider.getPatientEncounterVitalQuery()
+                .fetch("vital")
+                .where()
+                .eq("patient_encounter_id", encounterId)
+                .eq("vital.name", "weeks_pregnant")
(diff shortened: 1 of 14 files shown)


app/femr/business/helpers/QueryHelper.java
```diff
@@ -45,6 +45,21 @@ public static Float findPatientWeight(IRepository<IPatientEncounterVital> patien
         return weight;
     }
 
+    public static Integer findWeeksPregnant(IRepository<IPatientEncounterVital> patientEncounterVitalRepository, int encounterId){
+        Integer weeks_pregnant = null;
+        Query<PatientEncounterVital> query2 = QueryProvider.getPatientEncounterVitalQuery()
+                .fetch("vital")
+                .where()
+                .eq("patient_encounter_id", encounterId)
+                .eq("vital.name", "weeks_pregnant")
```
kevinzurek: Ken -

This should change to ".eq("vital.name", "weeksPregnant")"

The reason it wasn't immediately apparent is because the findWeeksPregnant(...) method is only called when retrieving a PatientItem and doesn't throw an error. You can see the problem if you create a new pregnant patient and try to see how many weeks pregnant they are on the medical page - the PatientItem will display a null value for weekspregnant.

The problem doesn't show up when viewing an old encounter because the history page uses a VitalMultiMap to get the value instead of a PatientItem.

-Kevin

kdunlap:
Good Find. I fixed it.

PR #447, opened 2015-11-21
Display Patient ID in Medical and Pharmacy FEMR-9

akrebs1202:
The Patient ID now displays in the patient overview within the medical and pharmacy section.

PR #452, opened 2015-11-23
Display Original Medications

AJSclyn:
Added functionality to display medications that were originally prescribed in medical, but they were changed in pharmacy.  The changes can be found in an encounter history as well as when a patient generates a pdf.


kevinzurek:
AJ -

It looks like we have password constraint commits in this pull request along with #449 . If you need help separating commits into different branches, take a look at the comment about cherry picking that i left Andrew:

https://github.com/FEMR/femr/pull/447

You can use a similar method to separate the password constraint commits from the display medication commits. I think that the same advice applies to #449 as the following commit exists there but not here:

https://github.com/AJSclyn/femr/commit/26b39a62f98f23ac4f4dc2199c7ccfd09fb3c7e4

-Kevin

PR #456, opened 2015-11-25
Pass Constraints with just pass constraint commits

AJSclyn:


app/femr/ui/views/triage/index.scala.html
```diff
@@ -228,7 +228,7 @@ <h2 class="text-center">Check In - Triage</h2>
                         Submit Patient</button>
                     <button type="button" class="fButton fOtherButton newPatientBtn pull-left">Reset Fields</button>
                 </div>
-
+                <div><h3>Submitting patient in history is for <b>new</b> encounters only.</h3></div>
```
kevinzurek: AJ,

I think this is from work on https://teamfemr.atlassian.net/browse/FEMR-22 and not the password constraint issue.

public/js/admin/users.js
```diff
@@ -86,22 +86,44 @@ var createUsers = {
...
             pass = false;
-        } else {
+        }
+        else if(!re.test(password)) {
+            passwordErrors +=  "password does not have a number ";
+            // createUsers.elements.password.next(".errors").text("password does not have a number\n");
+            pass = false;
+        }
+
+        if(pass == false)
```
kevinzurek: AJ,

You can use a strict comparison here. JavaScript uses both '==' and '==='. '===' won't match anything other than false. For example, this if statement will end up being true in the following scenarios while using the '==' operator:

pass = 0
pass = '0'

public/js/admin/users.js
```diff
@@ -86,22 +86,44 @@ var createUsers = {
...
-        $.each(document.forms["createForm"].elements["roles[]"], function () {
-            if ($(this).is(':checked')) {
-                isARoleChecked = true;
-            }
-        });
-        if (isARoleChecked === false) {
+        if (typeof document.forms["createForm"].elements["roles[]"] === 'undefined') {
             pass = false;
+        }
+        if (pass === false) {
```
kevinzurek: AJ,

It looks like the logic for validating if a role is checked has some flaws. I don't think the foreach loop (previously line 98-103) can be removed. Try the following steps to see what I am referring to:

1) Submit a new user with all valid fields, except no roles checked. The validation still passes and hits the server, which catches the error instead of catching it in JavaScript.

2) Submit a new user with no fields filled out. After the errors appear, select at least one role and click submit. The 'select at least one role' error no longer disappears.

kevinzurek:
AJ,

The formatting on the errors for creating a new user are not clearly displayed:

password

-Kevin

AJSclyn:
image

kevinzurek:
AJ,

Looks good when creating a new user and resetting a password.

I think the only other place a user can enter a password is in the edit user menu, but it doesn't seem to enforce constraints (/admin/users/edit/{userId})

-Kevin

kdunlap:
A couple of things I thought of after reviewing:

1. We should set the default date in the migration to a date in the past, not in the future.

change 9999-01-01 to something like 0000-01-01

Otherwise existing user accounts will never get flagged for password change. A downside I guess is that after this migration all user accounts will need to reset their password. Now that we have better password constraints, this is probably a good thing.


2. When logging in with a password older than 60 days, I get correctly redirect to the password reset screen, but I can immediately leave that screen and continue using femr without resetting the password. We should probably have some sort of check for the reset password flag at the router level so the user is not able to visit anywhere but that reset page until the password has been reset.

It may be a good idea to just add a subtask to this issue as something to tackle in the future though. Thoughts @kevinzurek?

AJSclyn:
So for checking the password on a routing level, I'm having a hard time understanding where to implement this.  On the password reset screen it looks like the only button the user can redirect themselves to is by clicking on the fEMR logo, which is the home page.  I was thinking just adding the same logic that I have in createPost() into createGet() in SessionsController.  Although if the person typed in a link like localhost/triage they would bypass the constraints again.

image

kevinzurek:
In my opinion, I think that solving the routing issue is for a separate task (or perhaps a sub task). I think the scope of this issue is just the password security itself. 

With that being said, I think that one solution for this is using an HttpFilter like we used to restrict Research access previously: 

https://github.com/FEMR/femr/blob/development/app/femr/util/filters/ResearchFilter.scala

Should we merge this pull request now?

kdunlap:
https://teamfemr.atlassian.net/browse/FEMR-76
https://teamfemr.atlassian.net/browse/FEMR-77

PR #457, opened 2015-11-25
Display Original Medications

AJSclyn:
Sections changed include Encounter History / PDF.

If the prescription was replaced in pharmacy, the encounter page will display the medication originally prescribed to the patient.

https://teamfemr.atlassian.net/browse/FEMR-15

### DIFF
--- a/app/femr/common/IItemModelMapper.java
+++ b/app/femr/common/IItemModelMapper.java
@@ -129,7 +129,7 @@ public interface IItemModelMapper {
      *
      * @param id            id of the prescription, not null
      * @param name          name of the prescription, not null
-     * @param replacementId id of the prescription that replaced this prescription, may be null
+     * @param originalMedications original prescription that replaced this prescription, may be null
      * @param firstName     first name of the person that prescribed the medication, may be null
      * @param lastName      last name of the person that prescribed the medication, may be null
      * @param medicationAdministration
@@ -137,7 +137,7 @@ public interface IItemModelMapper {
      * @param medication
      * @return a new PrescriptionItem or null if processing fails
      */
-    PrescriptionItem createPrescriptionItem(int id, String name, Integer replacementId, String firstName, String lastName,
+    PrescriptionItem createPrescriptionItem(int id, String name, IPatientPrescription originalMedications, String firstName, String lastName,
                                             IMedicationAdministration medicationAdministration, Integer amount, IMedication medication);
 
     /**
--- a/app/femr/common/IItemModelMapper.java
+++ b/app/femr/common/IItemModelMapper.java
@@ -130,7 +130,7 @@ public interface IItemModelMapper {
      *
      * @param id            id of the prescription, not null
      * @param name          name of the prescription, not null
-     * @param originalMedications original prescription that replaced this prescription, may be null
+     * @param originalMedicationName original prescription that replaced this prescription, may be null
      * @param firstName     first name of the person that prescribed the medication, may be null
      * @param lastName      last name of the person that prescribed the medication, may be null
      * @param medicationAdministration
@@ -138,7 +138,7 @@ public interface IItemModelMapper {
      * @param medication
      * @return a new PrescriptionItem or null if processing fails
      */
-    PrescriptionItem createPrescriptionItem(int id, String name, IPatientPrescription originalMedications, String firstName, String lastName,
+    PrescriptionItem createPrescriptionItem(int id, String name, String originalMedicationName, String firstName, String lastName,
                                             IMedicationAdministration medicationAdministration, Integer amount, IMedication medication);
 
     /**
--- a/app/femr/common/IItemModelMapper.java
+++ b/app/femr/common/IItemModelMapper.java
@@ -21,6 +21,7 @@ package femr.common;
 import femr.common.models.*;
 import femr.data.models.core.*;
 import femr.data.models.mysql.MedicationInventory;
+import org.joda.time.DateTime;
 
 import java.util.Date;
 import java.util.List;
--- a/app/femr/ui/views/history/indexEncounter.scala.html
+++ b/app/femr/ui/views/history/indexEncounter.scala.html
@@ -199,8 +199,8 @@
                         @for(x <- 1 to viewModelPharmacy.getPrescriptions.size) {
                             @defining(viewModelPharmacy.getPrescriptions.get(x - 1)) { prescription =>
                               <tr>
-                                  @if(prescription.getOriginalMedication != null) {
-                                      <th><p><del><span>@prescription.getOriginalMedication.getMedication.getName</span></del></p></th>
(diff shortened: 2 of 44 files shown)


kevinzurek:
AJ,

There is ambiguity in the UI around which prescription the original was replaced. For example:

original_dispensed


I can't tell which one replaced the original prescription.

-Kevin

AJSclyn:
image

app/femr/ui/views/history/indexEncounter.scala.html
```diff
@@ -177,15 +177,32 @@
...
 
-                    @for(prescription <- viewModelPharmacy.getPrescriptions){
-                        <p>
-                            <span class="infoLabel">Dispensed Medication:</span>
-                            @prescription.getAmount @outputStringOrNA(prescription.getName) (@outputStringOrNA(prescription.getMedicationForm).toString().trim)
-                            @for(activeDrug <- prescription.getMedicationActiveDrugs) {
-                                <span class='activeDrug'> @activeDrug.getName @activeDrug.getValue@activeDrug.getUnit </span>
-                            }
+                    @* AJ Saclayan Prescribed Medications *@
+                    @if(viewModelPharmacy.getOriginalMedications.size != null){
```
kevinzurek: AJ,

This is causing a warning from Java during compilation. The null check should exist to make sure you can call '.size()' on the original medications list. If 'viewModelPharmacy.getOriginalMedications()' is null, then you will get an error if you try to call .size().

The warning should go away if you remove .size().


This is the warning I get:
[warn] app/femr/ui/views/history/indexEncounter.scala.html:181: comparing values of types Int and Null using `!=' will always yield true
[warn]                     @if(viewModelPharmacy.getOriginalMedications.size != null){
[warn]                                                                       ^
[warn] one warning found

app/femr/ui/views/history/indexEncounter.scala.html
```diff
@@ -177,15 +177,32 @@
...
+                       <p> <span class="infoLabel">Originally Prescribed Medication:</span></p>
+                        @for(x <- 1 to viewModelPharmacy.getOriginalMedications.size -1) {
+@*//                           defining(viewModelPharmacy.getOriginalMedications.get(x - 1)) { medication =>*@
 
-                        </p>
+                            <p> <i>@viewModelPharmacy.getOriginalMedications.get(x-1).getName</i></p>
+
+                            <p> Replaced with: </p>
+                            <p>@viewModelPharmacy.getOriginalMedications.get(x-1 +1).getName</p>
```
kevinzurek: I think this just repeats a prescription. If i prescribe 3 medications (without dispensing/replacing any in pharmacy), this is what i get:

prescribed

Which makes sense because 'getOriginalMedications()' is just a list of prescribed medications that have the potential to be replaced (or are replaced) requested with 'searchService.retrieveReplacedPrescriptionItems(encounterId)' in the history controller.

I'm a bit concerned about repeating the dispensed medication twice (it took me a few to figure out what was going on) - once in the 'originally prescribed medication' section and again in the 'dispensed medication' section. Perhaps the sections could be combined and you could use strike-through text for prescriptions that were replaced? I'm open to ideas here.

AJSclyn:
Hey Kevin, 

Disregard my paint skills, but this is what me and Kevin planned on doing the layout for Prescriptions:

image

But the strikethrough idea seems cleaner.

Let me know.

kevinzurek:
I think your paint skills are great, but question - what is the difference between replaced, and dispensed?

AJSclyn:
You're right there is no difference.  I can implement a table with just Original and Replaced.  My question is if you can replace replaced meds?

app/femr/ui/views/history/indexEncounter.scala.html
AJSclyn: So adding just originalId into PrescriptionItem ( of course reflecting the changes where this needs to be used as well)?

kevinzurek: My thought was more along the lines of having a PrescriptionItem carry both the actual dispensed prescription and have it carry the original prescription, if needed. This would mean that wherever you need it you could just call something like "prescriptionItem.getDispensedPrescription()" and if it is a replaced prescription you could also call "prescriptionItem.getOriginalPrescription()" to get the original.

There's a mapper that is used to create all the ____Item objects, you can find the PrescriptionItem creation here:

https://github.com/FEMR/femr/blob/d3539e591d8d75fb2bb9aa363ff5e8ffccb614ff/app/femr/common/ItemModelMapper.java#L253

Of course, if this doesn't help you feel free to completely disregard!

AJSclyn: So in Prescription adding the originalPrescription name only, or originalPrescription <PrescriptionItem> within prescription item?  From my understanding too it looks like replacementId is never being set but only being passed as null.  Is this correct?

kevinzurek: I am referring to adding the originalPrescription within the prescription item, when available. I think it would make it easier to access since we don't have a service call to create the mapping of original prescriptions to replaced prescriptions.

Yeah, always being passed null seems to be the case. The description for the parameter actually reads: " id of the prescription that replaced this prescription, may be null". I think the original intent was to set the ID of the prescription that replaced the prescription, but I don't think a prescription that has been replaced is being retrieved anywhere in the application until now.

AJSclyn:
image
image

app/femr/common/IItemModelMapper.java
```diff
@@ -138,7 +138,7 @@ PatientItem createPatientItem(int id,
      * @param medication
      * @return a new PrescriptionItem or null if processing fails
      */
-    PrescriptionItem createPrescriptionItem(int id, String name, Integer replacementId, String firstName, String lastName,
+    PrescriptionItem createPrescriptionItem(int id, String name, String originalMedicationName, String firstName, String lastName,
```
kevinzurek: AJ,
Everything seems to work in the pull request. I am concerned that this change doesn't align with the rest of the code. If you were to use this change in ItemModelMapper, then the for loop you wrote in the SearchService (on line 352 - no idea how to link it here) could be completely removed. It would be replaced by sending the string into the stream on line 342 of SearchService. I think you could access the string for mapping more succinctly from the list of IPatientPrescriptions. Something like:

pp.getPatientPrescriptionReplacements().get(0).getOriginalPrescription().getMedication().getName()

It does work either way. I think if we keep the setting of item fields inside the ItemModelMapper it'd be easier to test and maintain later. 

-Kevin

kevinzurek:
this is not what i had in mind but it is definitely one way to do it!

kevinzurek:
rebased, squashed, and merged manually

AJSclyn:
:+1:  :+1:  :+1:  :+1:

PR #460, opened 2015-11-29
Time zones

zanni022:
[FEMR-56] Fix Stored Dates and Times

-There is a line added in the application.dev.conf file that I was not sure on how to make an evolution for. Please let me know at your earliest convenience and I can go ahead and make that change!

### DIFF
--- a/app/femr/util/calculations/dateUtils.java
+++ b/app/femr/util/calculations/dateUtils.java
@@ -21,12 +21,11 @@ package femr.util.calculations;
 import org.joda.time.*;
 import org.joda.time.format.DateTimeFormat;
 import org.joda.time.format.DateTimeFormatter;
+
 import java.text.DateFormat;
 import java.text.SimpleDateFormat;
 import java.util.Date;
-import java.util.Locale;
 import java.util.TimeZone;
-import java.util.logging.SimpleFormatter;
 
 /**
  * This class contains utilities for manipulating dates. If you add something here, please clearly document the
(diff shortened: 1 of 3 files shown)


kevinzurek:
See mailing list thread:

https://groups.google.com/forum/#!topic/team-femr/4rTjdDtDipI

kevinzurek:
Sam -

We are just waiting for the changes in the configuration files on this one.

-Kevin

app/femr/util/calculations/dateUtils.java
```diff
@@ -42,7 +43,7 @@ public static String getCurrentDateTimeString(){
     }
 
     public static DateTime getCurrentDateTime(){
-        return new DateTime();
+        return new DateTime(DateTimeZone.UTC);
```
kevinzurek: Sam -

Having trouble identifying the significance of this change along with line 100 after changing the JDBC connection string, is there one?

-Kevin

kevinzurek:
If I create a new encounter at 6pm EST before applying this change and create a new encounter at 6pm EST after applying this change. I think the database views both as UTC and when you view the first encounter it will say the patient was seen at 1pm and the last encounter will show 6pm.

Do we need to add 5 hours to every single time currently stored in the database before doing this?

app/femr/util/calculations/dateUtils.java
zanni022: Hi Kevin,

After the change that we made in the configuration file, I don't see the
use of this line anymore so I believe you are correct!

conf/application.conf
```diff
@@ -22,7 +22,7 @@ application.global="femr.util.startup.Global"
 
 #Register MySQL database settings
 db.default.driver="com.mysql.jdbc.Driver"
-db.default.url="jdbc:mysql://localhost/femr?characterEncoding=UTF-8"
+db.default.url="jdbc:mysql://localhost/femr2?characterEncoding=UTF-8&useLegacyDatetimeCode=false&serverTimezone=UTC"
```
kdunlap: make sure the connection string in application.conf and application.example.conf match the default database name of femr

`jdbc:mysql://localhost/femr?characterEncoding=UTF-8&useLegacyDatetimeCode=false&serverTimezone=UTC`

kdunlap:
To add to the comments issue of database times both before and after this change: we'll need to create something that will convert existing times in the database to UTC time. 

I think this PR solves the problem of storing all times in UTC, but we will probably need to solve the conversion of existing times before merging this code. We'll need to put some thought into this.

kevinzurek:
Re-opened here:

https://github.com/FEMR/femr/pull/474

PR #462, opened 2015-12-01
Added reference link to nav bar for med students

kdunlap:


kdunlap:
@kevinzurek: I'm looking at this again and I'm not sure I like the location of the reference icon in the nav bar. I'm going to do some thinking and might revisit this.

kevinzurek:
I'll merge it for now. If we improve it then another pull request can always be opened.

PR #463, opened 2015-12-01
Mark Patients as Deleted

AJSclyn:
Added new column for administrators to soft delete patients.  Patients that have been deleted will not be able to be searched in the Home Page, looked up in the Triage section, and the patient will not show up in the research data.

### DIFF
--- a/conf/evolutions/default/86.sql
+++ b/conf/evolutions/default/86.sql
@@ -6,4 +6,4 @@ ADD COLUMN `isDeleted` DATETIME NULL DEFAULT NULL AFTER `photo_id`;
 # --- !Downs
 
 ALTER TABLE `patients`;
-DROP COLUMN `isDeleted;
+DROP COLUMN `isDeleted`;
--- a/conf/evolutions/default/86.sql
+++ b/conf/evolutions/default/86.sql
@@ -1,0 +1,9 @@
+# --- !Ups
+
+ALTER TABLE `patients`
+ADD COLUMN `isDeleted` DATETIME NULL DEFAULT NULL AFTER `photo_id`;
+
+# --- !Downs
+
+ALTER TABLE `patients`;
+DROP COLUMN `isDeleted;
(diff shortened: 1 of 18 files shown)


kdunlap:
This code looks good. Just going to test to be sure.

kevinzurek:
This is great on the back-end, but how do i delete a patient as an Administrator?!

https://teamfemr.atlassian.net/browse/FEMR-36

I think we could accept this as long as we break FEMR-36 down into sub-tasks to indicate that the UI doesn't yet support this.

AJSclyn:
I can add a button when a patient looks at past encounter history that is only visible if the current user is an admin.  I was thinking right next to "View this Patient in Triage button".  Let me know before I implement the change.

kevinzurek:
I can dig it

AJSclyn:
image

conf/evolutions/default/86.sql
```diff
@@ -0,0 +1,9 @@
+# --- !Ups
+
+ALTER TABLE `patients`
+ADD COLUMN `isDeleted` DATETIME NULL DEFAULT NULL AFTER `photo_id`;
+
+# --- !Downs
+
+ALTER TABLE `patients`;
+DROP COLUMN `isDeleted;
```
kevinzurek: Just tried to execute the down script - 'isDeleted' needs a closing grave accent otherwise it won't work.

kevinzurek:
I'm a bit concerned that this doesn't allow someone who accidentally clicks delete to undo their action. A user can only click "Ok" and not "cancel".

AJSclyn:
image

kevinzurek:
I merged this manually

AJSclyn:
Yahoo ! :)

On Sat, Dec 5, 2015 at 8:01 PM, Kevin Zurek <notifications@github.com>
wrote:

> Closed #463 <https://github.com/FEMR/femr/pull/463>.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/FEMR/femr/pull/463#event-483582511>.
>



-- 
AJ Saclayan

PR #466, opened 2015-12-11
Cities Suggestive Typing

AJSclyn:
Not a perfect pull-request, but  I was wondering if I could get some assistance with the css for the typeahead dropdown because it's pretty much done.

On a full screen the type-ahead dropdown displays properly, but when I scale the website I get this : 
image
Typeahead injects css with their javascript, and I'm trying to override their css (triage.css).


kdunlap:
In regards to the css: In `triage.css` on line 354 the class `.tt-dropdown-menu` has a `left: 100px !important;` property. You need to override this when in mobile so that left is 0.

screen shot 2015-12-12 at 12 45 09 pm

screen shot 2015-12-12 at 12 46 57 pm

kdunlap:
A concern I have for this change is the loading time of the cities list. Our cities dictionary has 89,140 entries which the typeahead basically the entirety for its suggestion engine. The size of this query on my development machine is 1.8MB. 

screen shot 2015-12-12 at 12 43 56 pm

When fEMR is used in the field we always need to assume network conditions are at low range and slow. As it stands with this code now, every time the Triage page is loaded, the typeahead js makes a separate json request to get the list of cities, and downloads 1.8MB. I don't think this will be practical.

Quick solutions off the top of my head:
1. Limit the cities retrieved for the typeahead to those in the currently active trip
2. Don't grab any city suggestions until the user starts typing. This will take some extra js work on top of how the typeahead library is currently being used.

See https://twitter.github.io/typeahead.js/examples/ for examples. The Prefetch and Remote examples might be the most helpful

kevinzurek:
We could merge this with the assumption that there isn't a bajillion cities in the database, since the dictionary hasn't actually been added yet.

A new issue for importing the dictionary would have to be made - it would consist of narrowing the query in this pull request and the trip city dropdown for a superuser adding a trip.

kevinzurek:
merged manually

PR #469, opened 2015-12-22
FEMR118 - Can't view users on fresh install

kdunlap:
Changed retrieveAllUsers to return an empty set of users instead of an error when no users exist


kevinzurek:
merged manually

PR #479, opened 2016-01-28
[FEMR-155] Fix medical page exception

kdunlap:
Fixed issue with problems being submitted with fractional number indexes in the html form array


kevinzurek:
merged manually into master, 2.1, 2.2, and 2.3

PR #482, opened 2016-02-16
[FEMR-95] Add medications in medication inventory

AJSclyn:
Admins should be able to add medications to medication inventory based on the trip.

Notes: 
- Cherry-picked from femr-478.
- User must have a trip id assigned otherwise null error.

https://teamfemr.atlassian.net/browse/FEMR-95
&
https://teamfemr.atlassian.net/browse/FEMR-94

### DIFF
--- a/app/femr/ui/controllers/admin/InventoryController.java
+++ b/app/femr/ui/controllers/admin/InventoryController.java
@@ -75,7 +75,7 @@ public class InventoryController extends Controller {
         CurrentUser currentUser = sessionService.retrieveCurrentUserSession();
 
         InventoryViewModelGet viewModel = new InventoryViewModelGet();
-        ServiceResponse<List<MedicationItem>> medicationServiceResponse = medicationService.retrieveMedicationInventory();
+        ServiceResponse<List<MedicationItem>> medicationServiceResponse = medicationService.retrieveMedicationInventory(currentUser.getTripId());
         if (medicationServiceResponse.hasErrors()) {
             throw new RuntimeException();
         } else {
--- a/app/femr/business/services/core/IEncounterService.java
+++ b/app/femr/business/services/core/IEncounterService.java
@@ -30,11 +30,15 @@ public interface IEncounterService {
      * Create a new patient encounter. Chief complaint sort order is the same as
      * the order they exist in the list.
      *
-     * @param patientEncounterItem the new patient encounter, not null TODO: separate this into parameters.
+     * @param patientId id of the patient to create a new encounter for, not null
+     * @param userId id of the user creating the new encounter, not null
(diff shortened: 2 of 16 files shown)


kevinzurek:
Submitting the following input gives me an error:

error

kdunlap:
Just wanted to note that the issue with "old" medications kinda ties into: https://teamfemr.atlassian.net/browse/FEMR-97

Also, a possible way to fix the issue of being able to tell if a medication was from before the tracking addition or after would be to check if the medication "has" a medication_form relation.

**To get NEW medications**
Have the query for medications only grab those where the medication_form_id is not null

**To get OLD medications**
Have the query for medications only grab those where the medication_form_id is null

Just from a quick look at the schema, I think this will work.

kevinzurek:
Just to leave another note - I've finished most of this issue:

https://teamfemr.atlassian.net/browse/FEMR-96

I've marked it as blocked by FEMR95 and I'll PR it in after i rebase the changes from this pull request :+1:

AJSclyn:
Awaiting check :+1:

kevinzurek:
Seems good to me :+1: 

We should probably comment here:
https://github.com/FEMR/femr/blob/d9469b9785fb79d80760162501516b9abf0aeb8d/app/femr/ui/models/admin/inventory/InventoryViewModelPost.java#L48

and here:
https://github.com/AJSclyn/femr/blob/ba72c7169130ebaeada3920c09ebf05f189fd600/app/femr/business/services/system/MedicationService.java#L145

so that when someone decides a medication form isn't necessary in 5 years we will know the implications of removing it

app/femr/ui/controllers/admin/InventoryController.java
```diff
@@ -75,7 +75,7 @@ public Result manageGet() {
         CurrentUser currentUser = sessionService.retrieveCurrentUserSession();
 
         InventoryViewModelGet viewModel = new InventoryViewModelGet();
-        ServiceResponse<List<MedicationItem>> medicationServiceResponse = medicationService.retrieveMedicationInventory();
+        ServiceResponse<List<MedicationItem>> medicationServiceResponse = medicationService.retrieveMedicationInventory(currentUser.getTripId());
         if (medicationServiceResponse.hasErrors()) {
```
kdunlap: If a user is not assigned to a trip, currentUser.getTripId() is null, which Java does not like. This ties into https://teamfemr.atlassian.net/browse/FEMR-169

kdunlap: I'm going to merge this into master, then FEMR-169 will fix this issue

kdunlap:
Manually merged

PR #483, opened 2016-02-16
[FEMR-150] Don't let a user enter the same password that they currently have

btrvarun:
Don't let a user enter the same password that they currently have - solved

https://teamfemr.atlassian.net/browse/FEMR-150

### DIFF
--- a/app/femr/business/services/core/IUserService.java
+++ b/app/femr/business/services/core/IUserService.java
@@ -117,6 +117,12 @@ public interface IUserService {
      */
     ServiceResponse<IUser> update(IUser currentUser, Boolean isNewPassword);
 
+        /**
+     * Check if the old password is equal to the new password
+     *
+     * @param newpassword, the new password (not hashed), not null
+     * @param oldpassword, the old password (hashed), not null
+     * @return Boolean that returns true if the passwords are equal and false if not
+     */
     Boolean checkOldPassword(String newpassword,String oldpassword);
-    //Check if the old password is equal to the new password
 }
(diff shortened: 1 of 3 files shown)


app/femr/business/services/system/UserService.java
```diff
@@ -299,6 +298,11 @@ private void encryptAndSetUserPassword(IUser user) {
         user.setPassword(encryptedPassword);
     }
 
+    public Boolean checkOldPassword(String newpassword, String oldpassword)
```
kevinzurek: This needs the @override annotation to tell the compiler you are overriding a method in the IUserService interface (see other methods in the same class)

app/femr/business/services/core/IUserService.java
```diff
@@ -116,4 +116,7 @@
      * and/or errors if they exist.
      */
     ServiceResponse<IUser> update(IUser currentUser, Boolean isNewPassword);
+
+    Boolean checkOldPassword(String newpassword,String oldpassword);
+    //Check if the old password is equal to the new password
```
kevinzurek: If you use the javadoc template (see the comment on the update() method above) then you can click Ctrl+Q in IntelliJ while your cursor is on the method name to see what the method does.

btrvarun:
Added both corrections

kevinzurek:
Awesome! - @kdunlap i think we're good here

kdunlap:
This all looks good to me. Nice work @btrvarun!

kdunlap:
Manually merged

PR #484, opened 2016-02-17
[FEMR-120] track user deletes patient

ao3742:
femr120: “Track the user who deletes a patient”. When a patient is deleted,

the 'isDeleted' column is flagged in the patients table. A column should be added to track the id

of the user who deleted the patient. This will let us keep a history of who deletes a patient.

Above issue has been resolved in this code. 
Thanks.

### DIFF
--- a/app/femr/data/models/mysql/Patient.java
+++ b/app/femr/data/models/mysql/Patient.java
@@ -53,6 +53,9 @@ public class Patient implements IPatient {
     private List<PatientEncounter> patientEncounters;
     @Column(name = "isDeleted", nullable = true)
     private DateTime isDeleted;
+    @Column(name = "deleted_by_user_id", unique = false, nullable = false) 
+    private int deletedByUserId;
+
 
     @Override
     public int getId() {
@@ -163,5 +166,32 @@ public class Patient implements IPatient {
     public void setIsDeleted(DateTime isDeleted) {
         this.isDeleted = isDeleted;
     }
+/*
+    @Override
+    public int getDeletedByUserId() {
+        return deleted_by_user_id;
+    }
+
+    @Override
+    public void setDeletedByUserId(int deleted_by_user_id) {
+        this.deleted_by_user_id = deleted_by_user_id;
+    }
+    
+  */
+  
+    @Override
+    public int getDeletedByUserId() {
+        return deletedByUserId;
+    }
+
+    @Override
+    public void setDeletedByUserId(int userId) {
+        this.deletedByUserId= userId;
+    }
+  
+    
+    
+    
+    
 
 }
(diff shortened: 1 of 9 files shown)


kevinzurek:
@ao3742 did you close this on purpose?

app/femr/data/models/mysql/Patient.java
```diff
@@ -53,6 +53,9 @@
     private List<PatientEncounter> patientEncounters;
     @Column(name = "isDeleted", nullable = true)
     private DateTime isDeleted;
+    @Column(name = "deleted_by_user_id", unique = false, nullable = false)
```
kevinzurek: nullable should be true because the column always has a null value unless a patient has been deleted

conf/evolutions/default/87.sql
```diff
@@ -0,0 +1,9 @@
+# --- !Ups
+
+ALTER TABLE `patients`
+ADD COLUMN `deleted_by_user_id` INT(11) NULL DEFAULT NULL  AFTER `isDeleted`;
```
kevinzurek: a foreign key constraint should be added between this column and the id column in the users table. This ensures the ID always exists in the users table

conf/routes
```diff
@@ -56,7 +56,7 @@ GET         /photo/encounter/:id                               @femr.ui.controll
 GET         /triage/:id                                        @femr.ui.controllers.TriageController.indexPopulatedGet(id: Integer)
 GET         /triage                                            @femr.ui.controllers.TriageController.indexGet()
 POST        /triage                                            @femr.ui.controllers.TriageController.indexPost(id: Integer)
-POST        /triage/:id                                        @femr.ui.controllers.TriageController.deletePatientPost(id: Integer)
+POST        /triage/:id                                        @femr.ui.controllers.TriageController.deletePatientPost(id: Integer, deleteByUserID: Integer)
```
kevinzurek: Adding a route is one way to do this, but the deletePatientPost() method has a currentUser variable which will always contain the ID of the user (just use currentUser.getId())

kevinzurek: Adding a route is one way to do this, but the deletePatientPost() method has a currentUser variable which will always contain the ID of the user (just use currentUser.getId())

ao3742: Hi Kevin, I’ve modified my code per your request below and it is working now.

Thanks,
Vivek

app/femr/data/models/mysql/Patient.java
ao3742: Hi Kevin, I have updated my code, now nullable is true.

Thanks,
Vivek

conf/routes
kdunlap: Hey Vivek - @ao3742 

Glad you were able to implement @kevinzurek's changes. Make sure you commit those changes and push them to the branch on your fork to update the pull request. All you need to do is push to your remote's fork (origin) and this pull request will update with those new changes. Then we can merge. Let us know if we can help out with that. Good Luck!

kevinzurek:
replaced by #490

PR #487, opened 2016-03-02
[FEMR - 126] Add description to administrator settings

kratikaswami:


conf/evolutions/default/59.sql
```diff
@@ -7,6 +7,43 @@ CREATE TABLE `system_settings` (
...
+UPDATE system_settings SET description =
+'This setting must be selected if you want to allow the research only option. '
+WHERE id = 7;
+
+UPDATE system_settings SET description =
+'This setting must be selected to prompt for diabetes. '
+WHERE id = 8;
+
+
 # --- !Downs
```
kdunlap: Hey @kratikaswami. Everything looks pretty good here, just some small things to tweak up.

The above edit to Evolution 59 looks pretty good, but there are a few things to take care of:

1. These changes should be in their own evolution file. 

The idea is that any existing femr install can have its code base updated, and it will automatically run the evolutions to change the database to how the code will expect it to be. In this case, there are femr setups that are already past evolution 59. In order to get these new changes, those systems will have to run the downs on every evolution they've already run (we are at 86) down to evolution 59. When this process is done, evolution 59 through 86 will have to run their ups again. This will cause some data loss as existing tables and columns will have to be removed during the downs and re-created with no data in the ups back to 86.

All you need to do is create a new evolution and copy/past the new sql to the new evolution (87.sql).

2. Some of the descriptions in this file can also be tweaked. See the Jira issue for this task (https://teamfemr.atlassian.net/browse/FEMR-126) for brief descriptions of what the settings do.

Also, none of the settings are compulsory or required to be checked. Feel free to tweak the descriptions on the Jira task as you see fit and let us know if you have any questions regarding this.

app/femr/ui/views/admin/configure/manage.scala.html
```diff
@@ -13,13 +13,14 @@
...
+                <th>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Description</th>
             </tr>
 
             @for((key, valyew) <- viewModel.getSettings) {
                 <tr>
                     <td class="name">@key</td>
                     <td class="isActive">
-                        <input type="checkbox" name="settings[]" value="@key" @if(valyew) { checked="checked"} />
+                        <input type="checkbox" name="settings[]" value="@key" @if(valyew) { checked="checked"} >
                     </td>
```
kdunlap: Maybe this was taken out by accident, but the closing slash makes that input field valid XHTML. This is not the end of the world, but could you add that back in? Line 23 should read:

`<input type="checkbox" name="settings[]" value="@key" @if(valyew) { checked="checked"} />`

kratikaswami: Hi Ken,
Thanks for the review. I will make the required changes and submit again. I
added the "/" intentionally to the line 23 as it was showing an error, I
wasn't knowing that it would create other problems. I would remove it
anyways.

Thanks again,
Kratika

kevinzurek: Hi Kratika -

FYI - You don't need to resubmit if you commit your change and push it to your fork on the same branch as this (it will update automatically in this pull request).

Thanks!!

conf/evolutions/default/59.sql
kdunlap: Also - I haven't fully run this code yet, but a couple things:

1. I would check into adding the description field to your `models` and `items`. These work together pass the new field down from the data layer to the service layer into the controller and to display in the view. I don't think this is currently working - meaning, do the descriptions actually show up on page?

2. Make sure to add the downs section in the new migration which will drop the new description column.

kevinzurek:
not mergeable - closing

PR #488, opened 2016-03-02
[FEMR-157] Added phone number field to triage page

Ayush666:


app/femr/ui/views/triage/index.scala.html
```diff
@@ -45,10 +45,17 @@ <h2 class="text-center">Check In - Triage</h2>
                 @inputText("First Name", "firstName", true, if(viewModel != null) viewModel.getPatient.getFirstName else null, "text")
                 @inputText("Last Name", "lastName", true, if(viewModel != null) viewModel.getPatient.getLastName else null, "text")
                 @inputText("Address", "address", false, if(viewModel != null) viewModel.getPatient.getAddress else null, "text")
+
+
+                @inputText("Phone No.", "phoneNo", true, if(viewModel != null) viewModel.getPatient.getPhoneNo else null, "text")
```
kevinzurek: @Ayush666 change the third parameter to false because this field shouldn't be required (this should remove the red star in the user interface)

kevinzurek:
@Ayush666 This is also missing the evolution to create a new column in the patients table. Without the new evolution, the new column doesn't exist and i get an error when saving.

I've also added some additional JIRA tasks based on this feature being completed:
https://teamfemr.atlassian.net/browse/FEMR-194
https://teamfemr.atlassian.net/browse/FEMR-195

PR #489, opened 2016-03-03
[FEMR-183] implement front end medication ui

kdunlap:
https://teamfemr.atlassian.net/browse/FEMR-183

### DIFF
--- a/app/femr/common/ItemModelMapper.java
+++ b/app/femr/common/ItemModelMapper.java
@@ -21,10 +21,8 @@ package femr.common;
 import femr.business.helpers.LogicDoer;
 import femr.common.models.*;
 import femr.data.models.core.*;
-import femr.data.models.mysql.PatientPrescription;
 import femr.util.calculations.dateUtils;
 import femr.util.stringhelpers.StringUtils;
-import org.joda.time.DateTime;
 
 import java.util.Date;
 import java.util.List;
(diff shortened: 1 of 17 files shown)


kevinzurek:
@kdunlap code seems to be legit, the only requirement missing from the JIRA ticket is:

"when choosing an administration frequency and entering a number of days, the amount should be automatically calculated"

kdunlap:
Ok good find, nice work testing. The thing about that though, this issue is why I ended up creating FEMR-166 (https://teamfemr.atlassian.net/browse/FEMR-166) to fix the daily_modifier column.

Check your medication_administrations table in the database you used - are all the modifiers 0.00? The javascript that does the calculation relies on values that come from that table. Somehow an evolution (or something else) is setting those values to 0 after they get seeded initially.

I'm not sure if those values belong in the database or not though. I want to put some thought into alternative solutions.

kevinzurek:
Ah yes, I remember that now (I honestly have no idea how the modifiers are being set to 0 after the database seeder). I've done a few things:

1) Strike through the requirement in https://teamfemr.atlassian.net/browse/FEMR-183 that is automatically calculating the amount.

2) Added https://teamfemr.atlassian.net/browse/FEMR-197 which is the actual issue to automatically calculate the amount. 

3) I've marked FEMR197 as blocked by https://teamfemr.atlassian.net/browse/FEMR-166 so that we fix the modifiers in the database first.

4) Going to merge this.

@AJSclyn All the added tasks have been put in the medication tracking epic and added to the medication tracking sprint :+1:

PR #491, opened 2016-03-05
[FEMR-148] Inform the user of a possible match

btrvarun:
The system will detect a match if the first name, last name, city and date of birth are equal.

### DIFF
--- a/app/femr/ui/controllers/TriageController.java
+++ b/app/femr/ui/controllers/TriageController.java
@@ -3,26 +3,30 @@ package femr.ui.controllers;
 import com.google.gson.Gson;
 import com.google.gson.reflect.TypeToken;
 import com.google.inject.Inject;
-import femr.business.helpers.LogicDoer;
 import femr.business.services.core.*;
 import femr.common.dtos.CurrentUser;
 import femr.common.dtos.ServiceResponse;
-import femr.common.models.*;
+import femr.common.models.PatientEncounterItem;
+import femr.common.models.PatientItem;
+import femr.common.models.SettingItem;
+import femr.common.models.VitalItem;
 import femr.data.models.mysql.Roles;
 import femr.ui.helpers.security.AllowedRoles;
 import femr.ui.helpers.security.FEMRAuthenticated;
-import femr.ui.models.triage.*;
+import femr.ui.models.triage.EditViewModelPost;
+import femr.ui.models.triage.IndexViewModelGet;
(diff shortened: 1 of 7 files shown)


app/femr/ui/controllers/TriageController.java
```diff
@@ -78,14 +80,21 @@ public Result indexGet() {
...
         viewModelGet.setPossibleAgeClassifications(patientAgeClassificationsResponse.getResponseObject());
+    }
 
-        return ok(index.render(currentUser, viewModelGet));
+    public Result indexGet() {
+        //initalize an empty patient
+        PatientItem patientItem = new PatientItem();
+        CurrentUser currentUser = sessionService.retrieveCurrentUserSession();
+        IndexViewModelGet viewModelGet = new IndexViewModelGet();
+        indexGetInitialize(currentUser,patientItem,viewModelGet);
```
kevinzurek: what is the purpose of indexGetInitalize here and below (line 181)?

kevinzurek:
@btrvarun This is a really good start. The original intent of this issue was not to objectively say a patient has been matched and found (for example, what if multiple patients exist with the same name), but rather to inform the user that fEMR has detected a duplicate patient MAY exist. I did strikethrough the wording on the issue for FEMR148 to reflect this:
https://teamfemr.atlassian.net/browse/FEMR-148

To add more context - Eventually (and outside the scope of the issue you are currently working on) we will want the user to be able to say "Okay fEMR, i think these ARE the same patients" and then fEMR will merge the two entries:
https://teamfemr.atlassian.net/browse/FEMR-149

With that being said, as part of FEMR148, I think we should allow for some level of ambiguity. For example, lets say a user exists in the database as:
first name: Jacquiline
last name: Smith
age: 34
city: St. Jude

We will still want to inform the user of a possible match when they enter the following into Triage:
first name: Jacquileen
last name: Smith
age: 34
city: Port Au Prince

Note the slight difference in spelling of the first name and the alternate city. @AJSclyn used a PHP function that implemented an algorithm named "Levenshtein distance" here:

https://github.com/FEMR/fEMR-Cleanse-Generator/blob/350de7a2bbc2b8f7a4ce776aaa05bbd9ebf727e3/ADMIN%20PAGE/cityCleanse.php#L110

If you can find an implementation of this algorithm in a Java library - please import it!!

kevinzurek:
Closing this until it's ready to re open

PR #492, opened 2016-03-10
[FEMR96] subtract amount from inventory

kevinzurek:
Subtracts from the current quantity of a medication on dispensing of the prescription (clicking submit on pharmacy)

https://teamfemr.atlassian.net/browse/FEMR-96

### DIFF
--- a/app/femr/business/services/core/IInventoryService.java
+++ b/app/femr/business/services/core/IInventoryService.java
@@ -65,6 +65,7 @@ public interface IInventoryService {
      * @return a medication item that contains quantity information.
 
     ServiceResponse<MedicationItem> setQuantityCurrent(int medicationId, int tripId, int quantityCurrent);
+    */
 
     /**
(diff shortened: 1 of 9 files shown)


kdunlap:
Everything seems to work pretty well here. In testing I found that replaced medications do not get marked as dispensed. Seems to be a fairly quick fix.

- create an encounter and all that jazz
- prescribe a medication in the medical tab
- replace that medication in pharmacy
- the remaining quantity does not get decremented

kevinzurek:
@kdunlap good find. I ended up separating the action of "replacing", "dispensing", and "inventorying".

This actually uncovered another bug where it was thought that replacing was dispensing the medication as well, so replaced prescriptions were getting a date_dispensed of null.

Sorry for all the formatting changes in the diff :-1:

kevinzurek:
I also rebased this branch so re-checking out will be necessary

kdunlap:
Manually merged

PR #495, opened 2016-03-13
Task bug femr166 femr197 daily modifiers and prescription amounts

kevinzurek:
This is the issue with daily modifiers being turned to 0.00 after transitioning between evolution 75 and the newest.

It should fix both 

https://teamfemr.atlassian.net/browse/FEMR-166
and
https://teamfemr.atlassian.net/browse/FEMR-197


kdunlap:
Manually merged

PR #501, opened 2016-03-22
[Femr-148] Inform the user of a possible match

btrvarun:


app/femr/business/services/system/SearchService.java
```diff
@@ -775,4 +777,22 @@ private boolean isMetric() {
...
+     * @param first_name
+     * @param last_name
+     * @return true if the patient exists and false if not
+     */
+    @Override
+    public boolean checkExistingPatients(String first_name, String last_name, Date age,String city){
+        List<? extends IPatient> allPatients;
+        allPatients = QueryHelper.retrievePatients(patientRepository);
+        for (IPatient patient : allPatients) {
+        if ((patient.getFirstName().equals(first_name)) && (patient.getLastName().equals(last_name)) && (patient.getAge().equals(age)) && (patient.getCity().equals(city))) {
```
kevinzurek: this is throwing errors when patients don't have an age. To get a patient without an age, create a new patient and don't set age or birthdate, but check child or infant or adult.

I think it would be good to put this in a try{}catch{} statement, it would also help to check for null before using the age parameter

app/femr/business/services/core/IPatientService.java
```diff
@@ -51,7 +51,7 @@
      * @return a service response that contains a PatientItem representing the patient that was created
```
kevinzurek: a comment describing the check parameter should be added here so people know what it is doing in the future

btrvarun:
Added the changes

kevinzurek:
Closing for now... implementation performs a mass lookup query and then filters over a for loop. the database lookup should be constrained to first name/ last name/ etc

PR #500, opened 2016-03-22
[FEMR-157] Added phone number field to triage page

Ayush666:

PR #502, opened 2016-03-23
Improvement femr157 add phone number new

Ayush666:
Rebased and reopening https://github.com/FEMR/femr/pull/500 has issue with https://github.com/Ayush666/femr/blob/improvement-FEMR157-addPhoneNumber-New/app/femr/ui/views/partials/encounter/dynamicTabSpan.scala.html

### DIFF
--- a/app/femr/business/services/system/SearchService.java
+++ b/app/femr/business/services/system/SearchService.java
@@ -423,6 +423,7 @@ public class SearchService implements ISearchService {
         String firstName = null;
         String lastName = null;
         String firstOrLastName = null;
+        Integer phoneNo = null;
         if (words.length == 0) {
             //nothing was in the query
             response.addError("", "query string empty");
@@ -432,6 +433,7 @@ public class SearchService implements ISearchService {
             try {
                 //see if it is a number
                 id = Integer.parseInt(words[0]);
+                phoneNo = Integer.parseInt(words[0]);
             } catch (NumberFormatException ex) {
                 //see if it it a string
                 firstOrLastName = words[0];
@@ -453,7 +455,8 @@ public class SearchService implements ISearchService {
             //this is the most ideal scenario
             query = QueryProvider.getPatientQuery()
                     .where()
-                    .eq("id", id)
+                    .or(Expr.eq("id", id),
+                            Expr.eq("phone_no",phoneNo))
                     .isNull("isDeleted")
                     .order()
                     .desc("id");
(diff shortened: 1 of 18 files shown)


app/femr/ui/views/history/indexPatient.scala.html
```diff
@@ -97,7 +97,9 @@ <h3 class="pull-left bold margin-top-zero">General Info</h3>
                     <label class="inputLabel">Age:</label>
                     <input type="text" class="form-control input-sm pull-left" name="age" value="@viewModel.getPatientItem.getAge" readonly/>
                     <label class="inputLabel">Sex:</label>
-                    <input type="text" class="form-control input-sm pull-left" name="age" value="@viewModel.getPatientItem.getSex" readonly/>
+                    <input type="text" class="form-control input-sm pull-left" name="sex" value="@viewModel.getPatientItem.getSex" readonly/>
```
kevinzurek: good catch :+1:

kevinzurek:
Everything works, new ticket for the issue mentioned here:

https://teamfemr.atlassian.net/browse/FEMR-207

kdunlap:
This all looks pretty good. I did find a small issue though. When adding a new patient the phone number get stored and everything works great. However, when creating a new encounter for an existing patient the phone number is not updated. 

In the Triage Controller (line 161) there is this code:

```
        if (id == 0) {
            patientItem = populatePatientItem(viewModel, currentUser);
            patientServiceResponse = patientService.createPatient(patientItem);
        } else {
            patientServiceResponse = patientService.updateSex(id, viewModel.getSex());
        }
```

Maybe we should change the updateSex function to updatePatient? @kevinzurek and @AJSclyn Do you think this change might warrant a new issue and we can just merge this pull request as is? Making this change might be out of scope. There is already more than one issue being corrected here

kevinzurek:
I would vote merge with a new issue

Ayush666:
FEMR-147-Allow a user to search for a patient by phone number

app/femr/business/services/system/SearchService.java
```diff
@@ -423,6 +423,7 @@ public SearchService(IRepository<IDiagnosis> diagnosisRepository,
         String firstName = null;
         String lastName = null;
         String firstOrLastName = null;
+        Integer phoneNo = null;
         if (words.length == 0) {
```
kdunlap: The phone numbers are stores as a string and really should be queried via string matching. Look into using `like` with wildcards to match the phone numbers. The SQL query should probably end up looking like:

```
... where phone_no LIKE '%1234%'  ...
```
This search function is a good place to start, but I think FEMR-147 is specifically referencing the search box on the home page. You'll need to look into the javascript used for the home page search as well.


Also, this would be best to submit in its own pull request since it is a different issue.

PR #503, opened 2016-03-23
fEMR-146 Editing medication inventory

AJSclyn:
https://teamfemr.atlassian.net/browse/FEMR-146

Some things to consider:
css edits/variable naming


app/femr/business/services/system/MissionTripService.java
```diff
@@ -155,6 +155,9 @@ public MissionTripService(IRepository<IMissionCity> missionCityRepository,
     @Override
     public Optional<IMissionTrip> retrieveCurrentMissionTrip(int userId) {
 
+        //TODO should this be moved to just be identified whenever a CurrentUser is retrieved?
+        // This should be contained in the session service??
+
```
kevinzurek: If this became part of the sessionService/CurrentUser and we wanted to get the current trip for a user that is not logged in (think editing a user as an Administrator and wanting to know this information) then I think we would be out of luck? This doesn't happen right now though.

Otherwise you're right

app/femr/ui/controllers/admin/InventoryController.java
```diff
@@ -208,11 +208,26 @@ public Result ajaxGet() {
...
             throw new RuntimeException();
         }
         return ok("true");
     }
+
+
+    /**
+     * Alters medication based on submit.
+     */
+    public Result editPost(int medicationID, int quantity, int tripId) {
```
kevinzurek: This might be better off being named ajaxEdit to stay in line with ajaxDelete

I would see editGet/editPost being the method names for the comprehensive edit page

app/femr/ui/models/admin/inventory/MedicationInventoryViewModelPost.java
```diff
@@ -0,0 +1,8 @@
+//package femr.ui.models.admin.inventory;
+//
+///**
+// * Created by AJ on 3/15/2016.
+// */
+//public class MedicationInventoryViewModelPost {
+//    private Integer medicationQuantity
+//}
```
kevinzurek: love it, best file in fEMR

kevinzurek:
The code looks good, the following scenario is my concern:

As an administrator using medication inventory in-country, I realize that someone has been prescribing the wrong amounts of medication, so I want to fix the value of current quantity.

^when the administrator does this, it will also impact the Total Quantity, but the total quantity has not change (someone was just giving out the wrong amounts of medication throughout the clinic day)

app/femr/business/services/system/MissionTripService.java
kdunlap: I agree with @kevinzurek. It does make sense that maybe this function could be in the sessionService, but I think keeping it in the MissionTripService we have the ability to get the Mission Trip for any user. 

Maybe instead we could rename this function to `retrieveMissionTripForUser(int userId)` to be more clear on what it is doing?

public/js/admin/inventory.js
```diff
@@ -22,6 +22,82 @@ $(document).ready(function () {
...
+    var manageUsers = {
+        toggleUser: function (object) {
+            var id = $(object).attr('data-id');
+            var value = $(object).attr('value');
+            var tripId = $(object).attr('tripid');
+            $.ajax({
+                url: '/admin/inventory/edit/' + id + "/" + value + "/" + tripId,
+                type: 'POST',
+                data: {},
+                dataType: 'text',
```
kdunlap: Typically when posting some data to a web server, the data to be updated (in this case value) should not be passed in via the url.

It is better to do something like:

```
$.ajax({
         url: '/admin/inventory/edit/' + id + "/" + tripId,
         type: 'POST',
        data: {
              'value': value
        },
```

Having the id and trip_id in the url is fine here because these values aren't being updated.

kdunlap:
First of all, nice work on this pull request @AJSclyn 

To add to what Kevin said above:

The `Total Quantity` header could be renamed to `Initial Quantity`. In my opinion, once a medication inventory is entered this value really should not change. There are probably a ton of scenarios I am unaware of and I'm sure after this feature gets used we'll have to revisit and tweak it a bit.

I propose:

- Rename the Total Quantity table header to Initial Quantity
- When a Current Quantity is edited, only change the current quantity value and leave the initial quantity

In addition, I really like how the input box pops up when you double click it. Maybe it would be good to also add a small edit icon next to the number. Then its quickly apparent to a user that they can edit that value. This might warrant another issue to be visited in the future though. I'd say its pretty certain that after some feedback this section might change a little bit.

AJSclyn:
Makes sense.  Should this be a feature in the comprehensive page?

AJSclyn:
Take back what I said. Keeping initial quantity as is and maybe showing
total quantity in comprehensive page?

Also noted on edit button.

kevinzurek:
Initial quantity: amount the administrator enters first
Total quantity: equal to initial quantity until receiving(purchase/donation) additional medication
Current quantity: total quantity minus total dispensed amount

I think this is good if we make sure editing current quantity does not imply a change in the total quantity.

PR #509, opened 2016-03-29
FEMR 206 - pharmacy always shows no medications

kdunlap:
I had to change way more than I bargained for, and I couldn't help myself in fixing a data layer naming convention. This should be tested thoroughly - I will be testing again myself.

https://teamfemr.atlassian.net/browse/FEMR-206

### DIFF
--- a/app/femr/business/services/system/SearchService.java
+++ b/app/femr/business/services/system/SearchService.java
@@ -350,31 +350,6 @@ public class SearchService implements ISearchService {
                 prescriptionItems.add(item);
             }
 
-//            List<PrescriptionItem> prescriptionItems = patientPrescriptions
-//                    .stream()
-//                    .filter(pp -> pp.getPatientPrescriptionReplacements() == null || pp.getPatientPrescriptionReplacements().size() == 0)
-//                    .map(pp -> itemModelMapper.createPrescriptionItem(
-//
-//                        pp.getId(),
-//                        pp.getMedication().getName(),
-//                        null,
-//                        pp.getPhysician().getFirstName(),
-//                        pp.getPhysician().getLastName(),
-//                        pp.getMedicationAdministration(),
-//                        pp.getAmount(),
-//                        pp.getMedication(),
-//
-//                        pp.getMedication()
-//                            .getMedicationInventory()
-//                            .stream()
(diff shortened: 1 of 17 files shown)


app/femr/business/services/core/ISearchService.java
```diff
@@ -73,10 +73,11 @@
      * Find all prescriptions that have not been replaced. This does not imply they have been dispensed.
      *
      * @param encounterId id of the encounter
+     * @param missionTripId id of the trip
      * @return a service response that contains a list of PrescriptionItems
      * and/or errors if they exist.
      */
-    ServiceResponse<List<PrescriptionItem>> retrieveUnreplacedPrescriptionItems(int encounterId);
+    ServiceResponse<List<PrescriptionItem>> retrieveUnreplacedPrescriptionItems(int encounterId, int missionTripId);
```
kevinzurek: This is the only thing that stands out - can't we infer the mission trip id if we have the encounter id? (mission trip id is a column in the patient_encounters table)

app/femr/business/services/system/SearchService.java
```diff
@@ -313,17 +317,22 @@ public SearchService(IRepository<IDiagnosis> diagnosisRepository,
...
+                        pp.getMedicationAdministration(),
+                        pp.getAmount(),
+                        pp.getMedication(),
+                        pp.getMedication()
+                            .getMedicationInventory()
+                            .stream()
+                            .filter( i -> i.getMissionTrip().getId() == pp.getPatientEncounter().getMissionTrip().getId() )
+                            .findFirst()
+                            .get(),
+                        pp.isCounseled()
```
kdunlap: @kevinzurek: So when fetching the inventory for a certain medication, we get back the inventories from all trips in the system - Obviously we only want the inventory for the current user's trip. What I did above works, but I feel like there is a better way. Am I missing something?

kdunlap:
https://teamfemr.atlassian.net/browse/FEMR-206

kevinzurek:
Found a bug - not able to prescribe a medication that DOES NOT exist in the medication inventory feature then go back to the Medical or Pharmacy module.

1)Create new patient in Triage
2)Prescribe a medication that does not exist in inventory in Medical, click submit
3)Try to go back to the same patient in Medical to add another prescription

PR #510, opened 2016-03-30
bug FEMR-192: Fix issue of new encounters created by patient history …

cd2499:
…button in Triage page

https://teamfemr.atlassian.net/browse/FEMR-192

Seems the form element for the history button on the triage page is nested inside the overall form for the submit which is causing issues (ie creating a new encounter on history button click).

This change is moving the history button/form outside of the overall submit form

### DIFF
--- a/app/femr/ui/views/triage/index.scala.html
+++ b/app/femr/ui/views/triage/index.scala.html
@@ -4,6 +4,7 @@
 @import femr.ui.views.html.layouts.main
 @import femr.ui.views.html.partials.search
 @import femr.ui.controllers.routes.TriageController
+@import femr.ui.controllers.routes.HistoryController
 @import femr.ui.views.html.partials.triage._
 @import femr.ui.controllers.routes.SearchController
 @import femr.ui.views.html.partials.helpers.outputStringOrNA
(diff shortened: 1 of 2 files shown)


app/femr/ui/views/triage/index.scala.html
```diff
@@ -263,6 +250,19 @@ <h2 class="text-center">Check In - Triage</h2>
...
+            <button type="submit" id="triageSubmitBtn" class="fButton fRedButton fSubmitButton pull-right">
+                Submit Patient</button>
+            <button type="button" class="fButton fOtherButton newPatientBtn pull-left">Reset Fields</button>
+            @if(viewModel.getPatient.getId != 0) {
+                @helper.form(action = SearchController.handleSearch("history")) {
+                    <input type="text" class="hidden" name="patientSearchQuery" value="@viewModel.getPatient.getId"/>
+                    <button type="submit" id="historyButton" class="fButton fOtherButton pull-left">Patient History</button>
+                }
+            }
+
```
kdunlap: This code works pretty well, but a form within a form isn't valid HTML. My thoughts are that this button does not need to be a form element. We could potentially replace the second form with an anchor tag like so:

```
@if(viewModel.getPatient.getId != 0) {
     <a href="@HistoryController.indexPatientGet(viewModel.getPatient.getId.toString)" class="fButton fOtherButton pull-left">Patient History</a>
}
```

It will just take some minor css to make the anchor tag look like the buttons on the page. It shoud get pretty close using the same classes though.

kevinzurek: I think the anchor tag is a much more elegant solution.

@kdunlap are you talking about nested forms or multiple forms on one page?

cd2499: Def agree. Much cleaner and less confusing than my attempt. Thanks for the suggestion @kdunlap

cd2499:
Hi guys,
So I rebased master, made my changes, then rebased master again. Looks like that's the order of operations in the Contributing file. Then I pushed forced (with lease). Really hope I got that all right.

Do I need to open a new pull request or can I continue working off of this one?

Thanks, let me know how to proceed of if I need more updates.

kdunlap:
@cd2499 Looks like your new commits are on the pull request, so it should all be good. We will check it out again and merge if its ready. After a quick look I think we should be good though.

For future reference, once you open a pull request all you need to do is update the branch on your fork and the pull request will also update. Sounds like thats exactly what you did. Nice work :thumbsup:

cd2499:
Thanks for all your help @kdunlap & @kevinzurek ! Very much appreciate it

AJSclyn:
Merged manually!
https://teamfemr.atlassian.net/browse/FEMR-192

PR #511, opened 2016-03-31
FEMR-97 Only show available available medications in typeahead

kdunlap:


kdunlap:
https://teamfemr.atlassian.net/browse/FEMR-97

kdunlap:
Noting for later: If no tripId (users not assigned to a trip), don't filter by tripId

kevinzurek:
Something worth noting:

If there's no trip ID (even if there is a trip ID) and a user enters a medication outside of what is in the inventory, it will get logged as an old medication (pre-2.2.0) because it won't have a form. I think this means that it won't show up in typeahead which might actually be okay.

kdunlap:
I agree that medications entered outside of inventory should not show up in the typeahead. 

I guess it could be possible that prescription ends up being used for something that can't be inventoried (not really sure how -- I'm not a doctor though). These types of questions are probably things we can tweak after getting some feedback.

kdunlap:
Fixed minor issues with evolutions 95 and 87

PR #519, opened 2016-04-10
FEMR125, FEMR220, FEMR227, FEMR228 wrap up inventory

kevinzurek:
Sending 4 JIRA issues in this one because they require the order 125 -> 220 -> 227 -> 228. Breaking the 1 issue per pull request rule here, but I think it would have been more confusing to do all these separately because they depend on each other... Squashed the commits to make it less confusing

125 (import dictionary): Create a new database seeder and populate it with all of the medication concept dictionary. This requires checking to make sure values don't already exist in the database (because this gets executed every time the application starts).
https://teamfemr.atlassian.net/browse/FEMR-125

220 (select from dictionary): Use the select2 library on the inventory page to let a user add a bunch of medications from the dictionary at once to the actual formulary for a trip. When they do this, the values are added to the non-dictionary tables. A user can not enter the amount here, i try to make up for that in 228.
https://teamfemr.atlassian.net/browse/FEMR-220

227(bug editing current quantity): Fix a bug that wasn't letting users edit the current quantity of a medication if it was in a paginated page (DataTables API stuff)
https://teamfemr.atlassian.net/browse/FEMR-227

228 (editing on inventory part 2): Let a user set the initial quantity for all medications without any business rules in the inventory table. This was the best idea I could come up with for a fast way to add medicine to the inventory
https://teamfemr.atlassian.net/browse/FEMR-228


kdunlap:
Holy seeder Batman! That did not look fun to write.

I noticed a few things when testing, but and going to look further.

1. I feel like we can prevent the duplications in the concept_medications table created by seeding. See the attached screenshot
2. Adding a medication with a blank name (e.g. Aspirin) causes some issues - namely there are checks in both model mappers to see if the name isNullOrWhitespace. This is easily fixed, but I want to wrap my head around the problem space a bit.

I am going to keep looking, but just thought I'd note for future reference.

Also - nice work with the commit messages 👍

kevinzurek:
Here's a quick entity relationship to help explain what i was thinking for the schema. 
-Red is a concept table, but displayed to the user as a drop-down (they have no way to add new/custom values)
-The green tables are exactly the same. The bottom green tables are for concepts and used as lookup values to make suggestions to the user. The top green tables are for actually storing what the user picks/customizes.

concept_dictionary



@kdunlap The 2 points you made:
1) The duplicates exist because each row in the concept_medications table represents a unique medication. "Tylenol 325mg Acetaminophen" and "Tylenol 500mg Acetaminophen" is the case where you end up with Tylenol twice. This uniqueness has to happen somewhere, it could possibly be a join table if we want to get rid of those duplicates.

2) Good catch!!!

kevinzurek:
Amended the fixes for the duplicates (#1) to FEMR-228 commit (will have to re-pull down)

kevinzurek:
Fixes from last comment:

https://github.com/FEMR/femr/pull/519/files#diff-970a8c334998d81569fb11524d409436L97
https://github.com/FEMR/femr/pull/519/files#diff-ec4fe6e6ead3fe553b408476272fd3dfL204
https://github.com/FEMR/femr/pull/519/files#diff-f7dfce91571ce882ba01c93178036ec2L52
https://github.com/FEMR/femr/pull/519/files#diff-9957fa09c0b8ebaf8ae293c7303f958dL84

kdunlap:
Remove the null check in the ItemModelMapper (line 309) if you could and I think this is ready to merge.

The empty names (Aspirin) create a minor ambiguity problem with the prescription names printed on page when revisiting an encounter and on the pharmacy tab, but I think we can make a new issue for that. I'll create the new issue when this gets merged and will probably try to knock it out as well.

screen shot 2016-04-19 at 8 17 34 pm

kevinzurek:
Good finds - I took out the null check when creating a PrescriptionItem in ItemModelMapper. I think this actually needs a little more work that just removing that check though. This line should only have to call the prescription name:

https://github.com/FEMR/femr/blob/e8393f580113b365f4f3280c5baf737dffdc8b96/app/femr/ui/views/partials/medical/tabs/prescriptionRow.scala.html#L43

We have one place generating a standard medication name, but not one generating a standard prescription name. I'll do this separately though.

Anyway, commit was amended to FEMR-228 so you'll have to repull.

PR #521, opened 2016-04-19
When creating a new user, require a password match

ao3742:
Improvement to femr-159 has been completed.


app/femr/ui/views/admin/users/create.scala.html
```diff
@@ -28,7 +28,7 @@
             'class -> "fInput",
             'placeholder -> "Email Address",
             '_label -> "Email Address",
-            '_isRequired -> true
+            '_isequired -> true
```
kdunlap: I think this line was inadvertently changed. It should read:

```
'_isRequired -> true
```

kdunlap:
Outside of the minor typo, this looks good and works well for me.

kevinzurek:
I think this was re-opened here: https://github.com/FEMR/femr/pull/524...

ao3742:
Hi Kevin,

I thought I deployed correct version in Improvement femr159 new user require password match ver2<https://github.com/FEMR/femr/pull/524> #524

I will submit it again now.


Thanks,
Vivek

PR #522, opened 2016-04-20
improvement FEMR216: Updates to better track if user was prompted to …

cd2499:
…screen for diabetes and whether the screening was performed. Including a new bool column in patient_encounters table to help identify. If column in table is null, user was not prompted. Otherwise, if patient was screened, value will be true. If patient was not screened, value is false.

### DIFF
--- a/app/femr/business/services/core/IEncounterService.java
+++ b/app/femr/business/services/core/IEncounterService.java
@@ -114,6 +114,8 @@ public interface IEncounterService {
      *
      * @param encounterId id of the encounter for the patient, not null
      * @param userId id of the physician that screened the patient for diabetes, not null
+     * @param isScreened true if patient was screened for diabetes, false if user opted not to screen patient,
+     * null if user was never prompted to screen patient for diabetes
      * @return updated patient encounter item and/or errors if they exist, or null if errors
      */
     ServiceResponse<PatientEncounterItem> screenPatientForDiabetes(int encounterId, int userId, Boolean isScreened);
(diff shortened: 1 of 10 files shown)


kevinzurek:
Greg, 

Everything works. Very clean pull request by the way :+1:

app/femr/business/services/core/IEncounterService.java
```diff
@@ -116,7 +116,7 @@
      * @param userId id of the physician that screened the patient for diabetes, not null
      * @return updated patient encounter item and/or errors if they exist, or null if errors
      */
-    ServiceResponse<PatientEncounterItem> screenPatientForDiabetes(int encounterId, int userId);
+    ServiceResponse<PatientEncounterItem> screenPatientForDiabetes(int encounterId, int userId, Boolean isScreened);
```
kevinzurek: Above this method are comments that describe each parameter - we should do the same for "isScreened" as this isn't very intuitive for a future developer. These comments show up whenever you Ctrl-q this method inside IntelliJ.

kevinzurek:
@kdunlap @AJSclyn so you guys are aware, this is how the patient_encounters table looks for the diabetes columns after a few test runs:

image
image


Line 1: prompt turned on, prompt appeared, user said yes
Line 2: prompt turned on, prompt appeared, user said no
Line 3: prompt turned on, prompt appeared, user said yes
Line 4: prompt turned off, did not appear
Line 5: prompt turned on, did not appear

I feel like this should also be documented somewhere, but i am not sure where. Maybe in the patient_encounter model?

app/femr/business/services/core/IEncounterService.java
cd2499: Nice, did not know about ctrl-q. Cool stuff. ok, updated with comments. Thanks for catching that

kdunlap:
Greg ( @cd2499 ),

Seconded on the clean pull request. 👍

kdunlap:
https://teamfemr.atlassian.net/browse/FEMR-216

PR #525, opened 2016-04-21
FEMR-208

btrvarun:
encounter PDF not displaying amount of prescription dispensed

### DIFF
--- a/app/femr/business/services/system/SearchService.java
+++ b/app/femr/business/services/system/SearchService.java
@@ -371,37 +371,41 @@ public class SearchService implements ISearchService {
                 .eq("encounter_id", encounterId);
 
         try {
-            List<? extends IPatientPrescription> patientPrescriptions = patientPrescriptionRepository.find(query);
             List<PrescriptionItem> prescriptionItems = new ArrayList<>();
-            for( IPatientPrescription pp : patientPrescriptions ) {
-                if(pp.getPatientPrescriptionReplacements().isEmpty()) {
-                    prescriptionItems.add(itemModelMapper.createPrescriptionItem(pp.getId(),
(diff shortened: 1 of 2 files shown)


kevinzurek:
Hi Varun,

I am trying to get this to work. I dispensed this prescription:
image

It shows up in the encounter history with the amount:
image

But i can't get the amount to show up in the PDF:
image

btrvarun:
Changed as needed. I tried to check if the medicationform was "null" before printing in the pdf so that I can change that to "(N/A)" as in the history page. But it gave a pdf error as when I tried to use the .equals() function so the pdf just prints null. But the amount is printed correctly.

kevinzurek:
I see what you're saying. A user not not want to see the value "null" though. It doesn't make sense to most non-technical people! 

Also, It looks like were using the "full prescription name" in multiple places and assembling it together in multiple places:

https://github.com/FEMR/femr/pull/525/files#diff-52080ea042e90027dcb53ba820d4318eR456

https://github.com/FEMR/femr/blob/5bb025957311c49993ce9100b6e613e5af1ba60f/app/femr/ui/views/history/indexEncounter.scala.html#L204

https://github.com/FEMR/femr/blob/5bb025957311c49993ce9100b6e613e5af1ba60f/app/femr/ui/views/partials/medical/tabs/prescriptionRow.scala.html#L43

If we assemble the "full prescription name" once here:
https://github.com/FEMR/femr/blob/5bb025957311c49993ce9100b6e613e5af1ba60f/app/femr/common/ItemModelMapper.java#L303

then we could say "prescriptionItem.getFullPrescriptionName()" in all 3 places above.

kevinzurek:
not mergeable - closing

PR #530, opened 2016-05-04
[FEMR-117] Patient Age Classification Backend fixed so user does not …

chrislogo:
…enter in the wrong age classification


kevinzurek:
Throws exception when you enter classification without age on Triage - closing

PR #532, opened 2016-05-04
Made it not required to need to change password when admin is editing…

chrislogo:
… a user


kevinzurek:
I believe this is FEMR117

kevinzurek:
158*

https://teamfemr.atlassian.net/browse/FEMR-158

PR #533, opened 2016-05-04
[FEMR-179] Appened tri date to the drop down list options in the team…

chrislogo:
… and country views


kevinzurek:
merged as part of #541

PR #535, opened 2016-06-08
FEMR-234 Fix null pointer exception

cd2499:
https://teamfemr.atlassian.net/browse/FEMR-234

### DIFF
--- a/app/femr/business/services/system/MedicationService.java
+++ b/app/femr/business/services/system/MedicationService.java
@@ -346,7 +346,7 @@ public class MedicationService implements IMedicationService {
     public ServiceResponse<PrescriptionItem> createPrescription(int medicationId, Integer administrationId, int encounterId, int userId, int amount, String specialInstructions) {
 
         ServiceResponse<PrescriptionItem> response = new ServiceResponse<>();
-        if (administrationId == -1)
+        if (administrationId !=null && administrationId < 0)
             administrationId = null;
 
         try {
--- a/app/femr/business/services/system/MedicationService.java
+++ b/app/femr/business/services/system/MedicationService.java
@@ -346,7 +346,7 @@ public class MedicationService implements IMedicationService {
     public ServiceResponse<PrescriptionItem> createPrescription(int medicationId, Integer administrationId, int encounterId, int userId, int amount, String specialInstructions) {
 
         ServiceResponse<PrescriptionItem> response = new ServiceResponse<>();
-        if (administrationId !=null && administrationId < 0)
+        if (administrationId !=null && administrationId <= 0)
             administrationId = null;
 
         try {
@@ -395,6 +395,9 @@ public class MedicationService implements IMedicationService {
             response.addError("", "medicationName can't be null or empty");
             return response;
         }
+
+        if (administrationId !=null && administrationId <= 0)
+            administrationId = null;
 
         try {
 


kevinzurek:
How did you manage to recreate this? 

administrationID would have to get past this check:
https://github.com/FEMR/femr/blob/5bb025957311c49993ce9100b6e613e5af1ba60f/app/femr/ui/controllers/MedicalController.java#L379

then it would have to hit this check:
https://github.com/cd2499/femr/blob/c1f7eadf3cfb5cea63568ab16db127f2f129714f/app/femr/business/services/system/MedicationService.java#L349

But the only possible values for administrationID should be -1 or greater than 0:
adminid

cd2499:
I'm going to need to pull Ken in on this, we looked at it together. I spent Friday evening trying to recreate what we did but I was unable to. I agree, it should come over as -1 if nothing was selected. Really can't recall how we passed in a null unless we forced it in?

kevinzurek:
paging @kdunlap

kdunlap:
So when the administrationID comes into the controller as -1, it gets set to null here:
https://github.com/FEMR/femr/blob/5bb025957311c49993ce9100b6e613e5af1ba60f/app/femr/ui/controllers/MedicalController.java#L379

Prior to the changes in this PR, the MedicationService would repeat the check in the controller, but checking if the administrationID == null would throw the null pointer exception error here:
https://github.com/FEMR/femr/blob/master/app/femr/business/services/system/MedicationService.java#L349

Why this happens, I didn't know until we came across this bug and had to look it up. I don't Java very well apparently. A relevant link: http://stackoverflow.com/questions/3352791/why-comparing-integer-with-int-can-throw-nullpointerexception-in-java

Basically since we are comparing the Object Integer (null) with a primitive int -1, unboxing of the Object Integer occurs, and since it is null, a NullPointerException gets thrown.

I learned something today

cd2499:
Thanks Ken. So on the front end, if I left the option as "Select", would I have received an error screen upon saving or would the exception be hidden away in a log or something? I have no idea why I was unable to get that error.

AJSclyn:
I was able to produce it by just changing the code back to the original (administratorId == -1) and selecting "Select" as the administration.  Looks good on my end.

Java is strange.

kevinzurek:
I was finally able to reproduce this.

The administration ID comes in to the MedicalController as -1 and can take 2 routes depending on if the medication name exists in the medication inventory. Each route uses a different set of code. I was only testing with a medication name that does not exist in inventory.

Route 1 (medication exists):
1) set adminID = null
https://github.com/FEMR/femr/blob/5bb025957311c49993ce9100b6e613e5af1ba60f/app/femr/ui/controllers/MedicalController.java#L350

2) create new prescription. Error on the null object comparison:
https://github.com/FEMR/femr/blob/5bb025957311c49993ce9100b6e613e5af1ba60f/app/femr/business/services/system/MedicationService.java#L349

Route 2 (medication is manually typed in - it does not exist in inventory):
1) set adminID = null
https://github.com/FEMR/femr/blob/5bb025957311c49993ce9100b6e613e5af1ba60f/app/femr/ui/controllers/MedicalController.java#L379

2) create a new prescription AND medication. Success:
https://github.com/FEMR/femr/blob/5bb025957311c49993ce9100b6e613e5af1ba60f/app/femr/business/services/system/MedicationService.java#L393


Since i don't think administration is a required value for either field, i think the fix should be the same for both routes?

cd2499:
Yeahhhh, in my haste I was entering "new" medications. Since that route doesn't do a conditional check on the admin value, I guess it got through without error.

So, I should go ahead and update the Prescription And Medication route with a similar check?

Thanks all, sorry for the delay on this.

kdunlap:
@cd2499 I think it would be a good idea to enforce a non-negative or null administrationId around this line also. 
https://github.com/FEMR/femr/blob/5bb025957311c49993ce9100b6e613e5af1ba60f/app/femr/business/services/system/MedicationService.java#L393

I'd say go ahead and add it to your existing code. FYI: If you commit to your existing branch and push to  your fork, the pull request will update with any changes.

cd2499:
Thanks for the help guys, much appreciated @kdunlap @kevinzurek @AJSclyn

kevinzurek:
merged manually

PR #536, opened 2016-06-15
[FEMR-236] Fixed logic in outputHeightOrNA to prevent null pointer error

kdunlap:


kdunlap:
My answer is no, but do we need to check for a string of "null" like here:

https://github.com/kdunlap/femr/blob/ce6b24161affffa80b34d52281915468de9dfc56/app/femr/ui/views/partials/helpers/outputHeightOrNA.scala.html#L5

I feel like a check for straight up null is good enough. If the vital for height gets stored as "null", I feel like there is an issue another place in the code.

kevinzurek:
merged manually

https://teamfemr.atlassian.net/browse/FEMR-236

PR #540, opened 2016-07-03
[FEMR237] add evolution to align play_evolutions table

kevinzurek:
Found this while working on a bigger issue detailed in the comments here:

https://teamfemr.atlassian.net/browse/FEMR-237

Play Framework started generating the play_evolutions table with a different datatype so this will sync up all of our old databases.


kevinzurek:
merged manually

PR #543, opened 2016-10-07
[FEMR-137] Flag birthdates when they are estimated

siosilvar:
Estimated flag is currently only exposed to the researchers in the charts and CSV export. If there is another use case for this, let me know.


siosilvar:
wrong PR, wanted to put this on my own branch first

PR #545, opened 2016-10-16
[FEMR-241] Add inventory report (CSV export)

jcreigh:


kevinzurek:
Hi Jess - attempting to export Inventory from full database and receiving error below when i click the "Inventory" button under Admin:

inventory_export_exception


To reproduce, you may need to import the database file: 5-20-2016haiti_98_.sql. Experiencing the same exception with database that is not de-identified.

jcreigh:
Totally neglected to actually verify if the user was on a trip. Added a check, and disabled the button if they aren't.

kevinzurek:
Good to merge on my end. Thanks!!

kevinzurek:
I got some feedback on the export from Haiti before merging:

1) It wasn't immediately clear what "quantityCurrent" and "quantityInitial" meant. Maybe we can change this to something like "Starting quantity" and "Leftover quantity"?

2) Medication #8960 "Adoxa 100 mg Doxcycline (caps)" (see picture) has more quantityCurrent than quantityInitial. In theory this is impossible, but users can fully customize these fields. Separate issue!

kevinzurek:
merged manually https://teamfemr.atlassian.net/browse/FEMR-241

PR #546, opened 2016-10-17
Changes made to address FEMR-157 - Optional Phone Number Record

Thraxus:
Full change log available upon request (below is pretty telling though!).


app/femr/ui/views/history/indexPatient.scala.html
```diff
@@ -97,6 +97,8 @@ <h3 class="pull-left bold margin-top-zero">General Info</h3>
                     <input type="text" class="form-control input-sm pull-left" name="firstName" value="@viewModel.getPatientItem.getFirstName" readonly/>
                     <label class="inputLabel">Last Name:</label>
                     <input type="text" class="form-control input-sm pull-left" name="lastName" value="@viewModel.getPatientItem.getLastName" readonly/>
+                    <label class="inputLabel">Phone Number:</label>
+                    <input type="text" class="form-control input-sm pull-left" name="phoneNumber" value="@viewModel.getPatientItem.getPhoneNumber" readonly/>
```
jcreigh: I think this should be `type="tel"`, since then it'll show up as a phone numpad when entering on mobile devices.

Thraxus: I chose text to keep it flexible for the unknown countries it may be used in. Tel does make more sense on the outside, but it isn't as flexible.

kevinzurek: 1) Excellent job on the pull request :+1: 
2) The way i understand this is that "tel" would limit us to only numbers. "text" would let users input characters. 
3) Do we have a case where a phone number would need a character? 
4) Majority of users are using a tablet

jcreigh:
tel limits you to numbers, pound, plus, and (at least on my phone) normal
formatting characters like parentheses, hyphens, etc. I've never heard of a
modern phone number anywhere needing anything else.

Thraxus:
I'm not familiar with how every place in the world formats their phone numbers, or in the rules and regulations someone may or may not follow when creating them. 

Given that, and the intended audience for this project, I'd rather be safe than sorry and leave it a text field.

jcreigh:
Everywhere should follow the global standards, else it would make
international calling pretty much impossible for them. Just looking at
Haiti and Ecuador, they follow the guidelines. I think it's worth it, since
it's a pain to enter a phone number with a virtual keyboard.

Maybe have something in the configuration in the admin panel to turn it off
if really necessary?

Thraxus:
I'm fine with whatever Kevin wants. 

However, I'm still on the train to leave it as text. It's one extra key on most virtual keyboards to swap from alpha to numeric, and it still provides protection for any number of unknown scenarios where an unformatted field would be required/favorable.

kevinzurek:
I think this is good to merge after we restrict phone to a numerical value. 

Screenshot from iPad :

image

Thraxus:
Provided I didn't screw up the update (tried to fix the commit message (spelling is hard) and may have failed in an epic way) - the change has been made.

kevinzurek:
Not sure what you did, but i'm able to compile & run. The history input field was changed to "tel", but that field is readonly... I think you meant to change Triage

Thraxus:
Yeah, My mistake.  Unfortunately, Triage isn't as straight forward to change as History.

History uses standard HTML to control the view.  Triage uses Play's form helper for it's fields.  @inputtext isn't controlled in a standard HTML way - changing type here (assumed "text" at the end of the entry) only managed to break the layout.

So far it looks like I need to dive into the .js (femr/public/js/triage/*) and figure out what needs to be added / changed here to control input.  However, that's just a guess.  I'll try to get to this tomorrow (Thursday, 10/27/16), but it may have to wait until the weekend.

Thraxus:
The other option is I redo Triage in HTML5.  

I'll explore both options.

jcreigh:
Changing the `"text"` to `"tel"` was proper, which you can see if you actually the HTML of the triage page. It only broke the layout because the [triage's css rules only work on text and date fields](https://github.com/FEMR/femr/blob/fea58ccb92b4b442b3f26cc34b4f76460755f4c0/public/css/triage.css#L132). 

Just add `.generalInfoInput input[type="tel"]` to that list (there's a few places it needs done in the file), and it'll fix the layout.

Thraxus:
Breaking the layout is irrelevant, fixing that is easy. The problem is the field still allows all characters. If you read above, the Scala in Play doesn't support the tel tag from HTML5, so no constraints are upheld. To get the field to function properly is a larger change than just tossing a few lines of code on the triage page. 

Maybe I'm missing something, but that's what I see at this point.

jcreigh:
I don't think it really matters if they enter invalid characters. The tel field is just a text field with context, so that it shows up as a phone keypad on mobile devices. Those constraints don't actually do any server side checking, so you won't be able to prevent anyone from entering random garbage anyway by editing the POST parameters directly. The backend already doesn't cleanly handle garbage on other fields. You can set the sex to something random and it'll happily put it in the database, or you can send text on a number field and it'll crash and hang the connection.

If you really want, you can prevent non-phone keys from being pressed with some javascript, no need to touch anything Play related. But personally, I wouldn't bother.

Thraxus:
The conversation we had with Kevin last week explicitly covered limiting what people can use the field for. 

Like I said way back in the beginning, I'm all for leaving it a text field - I don't believe saving one key press to swap to a tel field is/was that important. So if this change was just for that reason, I'd argue against the necessity of the required code changes given the new information. However, since the decision was also based on the field contents, the changes will need to be made. 

I have some ideas how to do it, and editing the js is one of them. The changes I will end up making all depend on how robust of a system to put in place. If I decide on just some regex validation, js will work fine. To upgrade the system as a whole and add missing functionality will do the system a greater service in the long run though. It's a decision that needs to be made with more information than I have available right now. I'll start looking into it tonight and let this thread know my thoughts before committing to anything.

jcreigh:
Oh, I don't remember that _at all_. I thought we were just talking about changing the field to tel so that on mobile devices the phone keypad will show up, not anything about actual validation. 

My opinion is that it should be tel, but shouldn't have validation. If @kevinzurek wants validation though, then fine. But that gets complicated, like you said. (Especially validation on the backend).

Thraxus:
Yeah, the wording was something similar to, 'it's best to limit what the field could be used for, so only allowing phone numbers is probably best, no one should be able to use it for notes or other random things'.

kevinzurek:
I think both points are important:
1) Save the extra click to switch to a numeric keyboard and
2) Restrict input to numerical values only

Without server validation (and assuming type="tel" is used), people using browsers with non-HTML5 or people that send HTTP requests directly would be able to persist non numerical values. I think this is acceptable since implementing Play server-side validation is a considerable amount of work in Triage.

@Thraxus it doesn't look like Play has support for this even in 2.4+. Wouldn't the quickest solution be to scrap using the @inputtext helper and write the html manually ?

Good discussion 👍

Thraxus:
I don't think any version of Play supports HTML5. The plugin I linked above was made for Play 2.5, so it's at least not included up to that point. 

My plan is to employ the best option available that requires the least disruption to the code in general. Translating the page from Play to HTML5 / CSS is the most likely course. However, I want to explore the changes that plugin makes before making a decision. Since I haven't had a chance to review it yet, for all I know it's only adding a few lines of code and we're golden. 

I'll post my findings here along with a plan before committing to anything. I'd rather we all be on the same page before starting I start coding.

Thraxus:
Alright, so, the only real option here is without massive amounts of work is to write out the Play form helper on the Triage page. 

This will require:
- Completely redoing the left portion of the Triage page in HTML5 / CSS / JS
- Refactoring the entire age selection area (possibly including the whole system regarding age selection [SQL, services, etc.])

The Age Selection Area relies completely on the Play Framework / View Model to function, so getting it to style properly with the rest of the page in HTML is either not possible without some major CSS work, or beyond my skill set (web design / CSS is not my forte).  To be fair, the age selection radio button section looks awful anyhow, so it could use a face lift.

Once this is done I'll be able to work on validation (likely just js here).

kevinzurek:
Hi Martin - i am not quite clear what you mean by "write out the Play form helper on the Triage page". You do have the option of building custom helpers like this one:

https://github.com/FEMR/femr/blob/master/app/femr/ui/views/partials/helpers/outputHeightOrNA.scala.html

I think this would be relatively quick. If the browser can read HTML5, then you can write HTML5 in a custom helper and it should work.

A lot of things need a face lift and I certainly support refactoring the age selection process as there are indeed some issues with it. At the same time, I would like for us to release a new version of fEMR before November 20th for December trips. Having the ability to save (and search? :)) phone numbers would be huge.

Thraxus:
Hi Kevin, 

What I mean is to remove all references to Play, and redo the page in HTML5 / CSS with regards to all data capturing.  Not only will this be an upgrade to the form as a whole, but it will also make future modifications a lot easier to deal with.  Writing a patch in the form of a custom helper isn't a bad idea, but I feel it's more of a band-aid than a true solution.  

If I'm not almost done with this fix by the weekend (as in, page layout complete at most of the data pipes reconnected), I'll change direction and go the helper route.  

I have high confidence I'll be able to integrate all page changes (and likely the search option!) well within your window for a November 20th release.

jcreigh:
I'm confused. The page pretty much _is_ HTML5 and CSS. It just uses the Play framework's Scala templating, which has helpers to generate forms (among other things).

I don't see why this specific thing can't be done by simply changing the type to `tel`, which is trivial, and by using Javascript to make sure that only phone number characters are entered. It'd only be client-side, and server-side validation would require quite a bit of rework in many places, but I think that's out of scope for this.

Thraxus:
No one has ever claimed these changes to be hard, simply time consuming. The most complicated part is adding in the validation, and I'll likely just keep that in the java script for now. 

The only issue here is time, which is in short supply with me. I work 60 hours a week, on average, Monday through Saturday, as well as take 8 credit hours worth of classes, on top of the myriad of other nuances life throws my way. If I had nothing to worry about but school, this would have been long since complete. 

Maybe I'm going about this a bit more complicated than is necessary, but as I mentioned above, I believe the project will be benefit from it overall. Play is nice, but a uniform design behind the pages means easier integration of any number of changes down the road. 

Kevin doesn't appear to mind the direction I'm taking. Unless he says otherwise, I'll stick to my plan.

kevinzurek:
Hi @Thraxus , what is the consensus here? Any progress? Implementation does not matter to me here as long as everyone thinks it's the optimal solution. It would be good to share your progress on this so we can give feedback as you work. 

Here's an example of the server side form validation i mentioned previously from play framework that uses these helpers:
https://github.com/FEMR/femr/blob/8986e4f540e480fef3318cb783bac2c9e7183d2a/app/femr/ui/models/admin/users/EditViewModel.java#L41

You have a great pull request here as it exists. The only missing piece is the "tel" type on the input element in Triage index.


Thanks!!

Thraxus:
Hi Kevin, 

As much as I'm sure you wanted to hear this, I ended up scrapping my changes last night. Doing the work the way I intended proved to require far more changes than originally anticipated. I'll be editing the helpers tonight to do the change. Complete code should be available for review tomorrow evening at the latest. 

Marty

Thraxus:
Update is incorporated without validation right now.  While I do think consistency is key across all forms, these changes were much easier (like 5 or 6 edits, as Jess mentioned) than what I was attempting (before I abandoned the project, I easily had over 300 edits in multiple places).  

Once I figure out all the hooks needed for the validation, I'll upload the code for review.  Again, target is no later than tomorrow night.

As a note, the only real validation I can put on the phone number is to restrict characters from the field.  The format used by every country in the world would require a RegEx filter that would take forever to assemble, and even longer to validate.  

See:
https://en.wikipedia.org/wiki/E.164
https://en.wikipedia.org/wiki/List_of_international_call_prefixes

... for the basic principals.  There are addon's that can help, but nothing I found that is a direct implementation.  

The tel field helps some when on a mobile device (the keypad is restricted to only valid key's), but is does nothing for the format or for entries done on a laptop / computer.  

An example of the poor validation on mobile devices, the following phone number is acceptable (using my phone as a test): ",,,,01N++();###N///-+,".  I could also restrict special characters, but then you run into the issue of a phone number being invalidated because someone accidentally hit a comma.  If the idea here is speed (which was the basis for this entire change before validation came into the equation), then I don't think restricting anything that's available on the mobile keypad would do anyone any good.    

One last note - looking at the data in the server on a new patient push, the age buttons (infant, child, teen, adult, elder) don't appear to push any data into the database.  I'm on a fresh copy of FEMR-157 (the code on this pull), so I don't believe this is an issue on my end.  Can someone else confirm?  I'd hate to push a screwed update.

Thraxus:
Just FYI, this is as far as the Validation has gotten so far:

2016-11-10 00_51_05-execution exception

I'd like to make that a tad more friendly though, so need to implement before a submit is accepted (same as with the 'required fields' rejection).  At least it stops data from hitting the database though, so it's a start.

jcreigh:
If we're going to include phone number validation we should _not_ write our own because we will _not_ do it right, since it entirely depends on locale. It's trivial to limit the characters to a standard character set, it's non-trivial to do anything regarding formatting.

So here's Google's i18n phone number library.

https://github.com/googlei18n/libphonenumber

It just needs the 2 character country code, which we can get from the trip country. There's no standard way to get this that I know of, but if we're going to assume English country names, then we can just [generate a map](http://stackoverflow.com/a/14155186) and look it up.

This'll also let us format phone numbers in the database in a standard way, which will let you search for them in the future. The javascript version is a Closure library, which we don't use but could probably figure out. The library also has a "AsYouTypeFormatter", if you want to make the phone entry field extra fancy.

See the demo (which is running Java in the backend).
http://libphonenumber.appspot.com/phonenumberparser?number=34567890&country=HT

Thraxus:
Unless I did something screwy, the change should be done.  I added a small JS hack to basically erase anything but a number form the tel field, and kept the back end check in place just in case something manages to get through.  It's the best we can do for validation without a ton of extra effort (and personally, I do not feel the extra effort is worth the minor gain; I'd rather direct that effort to other areas than need the attention more). 

I'll work on adding the phone number search on a different branch.

jcreigh:
You did something wonky. After your rebase you should have forced your push instead of attempting to merge. Now you have duplicates of commits with different hashes in different places in history.

I'd reset your branch to the previous commit, 1a2fe24, and then just force push that into your github. That should fix the wonkiness with the history.

app/femr/ui/models/triage/IndexViewModelPost.java
```diff
@@ -55,6 +58,13 @@
     //indicates if the "yes" button was clicked for the diabetes screening prompt
     private String isDiabetesScreenPerformed;
 
+    public List<ValidationError> validate(){
```
jcreigh: This causes a global error instead of handling it gracefully, which is better than nothing because it doesn't end up in the database, I guess, but since it's triggerable without much difficulty then I think it's an issue.

It also only checks against letters, whereas the regex client-side removes the first non-digit. Meaning you can send any symbol and it'll gladly be put into the database.

Thraxus:
I'll see what I can do about the pasting or holding a key, didnt think about those. Then I'll tackle your suggestion on fixing the push. Will update again in a bit.

Thraxus:
The global was the intent there - I didn't think it was easy to get to that point until your comment above. I'll update the regex to match the client though.

Thraxus:
Yeah, me and git aren't friends anymore.  055187181379e54aaf2b02bdff6ce41e6634f9c2 wasn't intended, but it has the same code as 3bbd008d787d4754915427248ceb22c81cf9419f just merged again. 

I added the global switch to the JS validation to remove all non-digits.  It should work cleanly on copy/paste, single key press, or a held down key now.   I also carried over the regex to the global fail.  

With the new client validation, I don't think it should be possible to hit the server, but I'm sure I didn't think of something.

kevinzurek:
@Thraxus The implementation is excellent - we need to get you and Git together to be friends though. I tried to clean it up myself, but i don't want to assume i got everything right:

https://github.com/kevinzurek/femr/commits/pr546_cleaned

If you are struggling with rebasing, you can always start a new branch off master and `git cherry-pick` in commits as needed. Otherwise we can review Git tactics Tuesday and get this PR merged.

kevinzurek:
@Thraxus still not friends with Git???

Thraxus:
@kevinzurek Nope, still not friends.  But we're working on it. 

Please scrap this request.  I've added a cleaned up (and improved) version under a new pull request.  I'll have the search option available tomorrow night, still having a small issue with the typeahead searches.

jcreigh:
No need to scrap it, you can just force push your new branch over your old remote one and it'll update this request.

jcreigh:
By the way, what do you think about having a checkbox or something for when the person searching wants to input a phone number instead of an ID? You can detect if it's pressed via javascript and adjust the form accordingly.

Something unintrusive like: https://jsbin.com/losezezeso

Just a thought.

Thraxus:
Too late on that one, I had already submitted the new pull request before making this push.  See pull 560. Plus, you know what happened last time I tried to get all fancy with this git stuff.  Baby steps!

As for the search option, I think that's a good idea if we stick with the basic search bar.  For the first run I think I'll just toss the phone number in to get a search working in general though.  We can improve upon it from there.

However, Kevin mentioned doing another change after to unify all the search types.  As in, the home page uses a typeahead format that doesn't use the same framework as the search bar on the other pages.  The goal with this change would be to make all pages use the typeahead so you get real time results.  When I get that working as intended, there should be no reason to differentiate search options because you'll get real time results based on what you've typed in (that's another problem with the search bar - there is no wildcard search, it's all or nothing).

Thraxus:
So as fun as it sounds, apparently I was the problem I was having with typeahead.  The improvements I did with 157 on the new pull implemented the typeahead integration without a problem.  Pull 561 only changes the field names to include the addition of the new search parameter.  Home still uses typeahead while Triage uses exact match.  

If any problems are seen on these, please let me know.

kevinzurek:
Closing this out, it shows up as referenced in #560

PR #549, opened 2016-10-23
[FEMR-246] Add Manager role to database and Roles class

jcreigh:
Simply adds the Manager role, nothing uses it yet.

### DIFF
--- a/app/femr/util/startup/DatabaseSeeder.java
+++ b/app/femr/util/startup/DatabaseSeeder.java
@@ -72,6 +72,7 @@ public class DatabaseSeeder {
         seedDefaultTabFields();
         seedPatientAgeClassification();
         seedDiagnosis();
+        seedUserRoles();
     }
 
     private void seedDiagnosis() {
@@ -995,6 +996,15 @@ public class DatabaseSeeder {
         return false;
     }
 
+    private static boolean containRole(List<? extends IRole> roles, String roleName) {
+      for (IRole role : roles) {
+        if (role.getName().equals(roleName)) {
+          return true;
+        }
+      }
+      return false;
+    }
+
 
     /**
      * Seed the admin user from the configuration file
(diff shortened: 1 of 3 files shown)


AJSclyn:
Hi Jessica,

It works when assigning the role to a user in the admin page.  Looks good to merge.  👍

kevinzurek:
Merged manually

kevinzurek:
Jess - 

Reopening this after merging. Found an issue if you:
1) update to this commit and the Manager role gets added to the system
2) Assign yourself the role
3) revert back to before this commit and apply the "down" evolutions

then the down evolution fails because of the foreign key constraint being enforced between the user from (2) and the roles table.

We added roles through evolutions a very long time ago. I think going forward that the evolutions should be used for DDLs only and the database seeder should be used for DMLs (this). I don't think it's safe to remove the Manager role after it's already been added.

jcreigh:
Is this satisfactory? Removed the evolution and added a role seeder to the DatabaseSeeder class.

app/femr/util/startup/DatabaseSeeder.java
```diff
@@ -1048,4 +1058,16 @@ private void seedAdminUser() {
...
         }
     }
+
+    private void seedUserRoles() {
+      // Other roles are in database from early evolutions
+      List<? extends IRole> roles = roleRepository.findAll(Role.class);
+      if (!containRole(roles, "Manager")) {
+        // Manager role doesn't exist, add it
+        Role newRole = new Role();
+        newRole.setId(Roles.MANAGER);
```
kevinzurek: Any reason for setting the primary key here? I am not sure what the behavior for this would be if a primary key with 7 already exists.

jcreigh: Yeah, I added it so it'll fail noisily instead of silently. Roles.MANAGER is hard coded to 7 in Roles.java, so if some other role (with an id of 7) exists for some reason, everything using Roles.MANAGER won't work properly.

But I can remove it if you don't see a point.

kevinzurek: Nope, you're right. This needs to be enforced. :+1:

kevinzurek:
Merged the fix!

PR #550, opened 2016-10-24
[FEMR-148] Alerts user if matching patient info already exists in database

C-RICHMOND:
After the user is alerted, they can choose to view the matching patient info and will be rerouted to the duplicate patient page.


AJSclyn:
Hiya,

Good work here.  Couple things after testing:

1.  I was only able to see the alert once.  Afterwards I clicked cancel and created a new patient with the same name, and I was unable to see the alert.  This has happened for other names as well.  @kevinzurek @kdunlap Any ideas why this is happening?
2.  Although this is done in javascript, did we want this logic done without javascript?

C-RICHMOND:
I'm not sure why the alert is only happening once, but I will look into it. Also, I had initially verified if their were any matching patients in the triage controller before a new patient is created, but later decided to implement the check in the javascript file instead. This way it is easier to alert the user based on the results of the returned JSON object. That being said, you would rather I implement it in the backend?

C-RICHMOND:
@AJSclyn @kevinzurek 
From what I understand of FEMR-148, the objective is to alert the user that there is an existing patient with the same first and last name in the database. This is simply just an alert, and the patient will still be created regardless of the action the user takes on the alert. If the the user confirms the alert by clicking okay, they will be redirected to the duplicate patient page, which will show all the instances of matching patients, including the newly created patient. If they cancel, the patient will be created and they will be routed as normal.

In regards to implementing the change in javascript, all I am doing in javascript is alerting the user if there is a match in the database, and then rerouting the user to the duplicate patient page. The real logic happens in an existing function, doesPatientExist(String query), in the Search Controller. I make this API endpoint call and handle the result in javascript, but the logic is implemented in that controller.

Finally, addressing the issue you encountered. After submitting a patient, getting an alert, and then clicking cancel, the user is directed to the patient history page where it displays the patient Id and applicable info. If the user wishes to create a new patient, they have to go back into Triage and fill out the information. If the names match, the user will again be alerted and can decide what to do. I have tested this multiple times and it seems to be working.

Sorry for the long comment, but please reply if you have any questions or I need to make any code modifications.

kevinzurek:
Hi Chris - you are correct that this is just an alert and that the patient should be created regardless. It's good that you are using the same logic that already exists in the SearchController - nicely done. I don't see an issue with calling this in JavaScript. 

I need to pull this down and test it still in regards to the issue that @AJSclyn mentioned.

In the future we will need to figure out how to "de-dupe" these records.

kevinzurek:
@C-RICHMOND I was about to merge this, but noticed at the last second that it seems like photos aren't uploading. When i submit a photo on the master branch it works, but not here. Still trying to figure out why.

I also changed the text for the prompt slightly so the user knows their data will still be persisted:
https://github.com/kevinzurek/femr/commit/073d0bd8c78c1866e2d8c017af52fc5636ba0c2d


patientPhotoFeature.prepareForPOST();

this actually removes the photo from the DOM and should be called regardless of a match. Photos get compressed on the client before they get submitted to the server via HTTP because users often have issues with bandwidth.

C-RICHMOND:
@kevinzurek I made a few changes to the code structure and committed. After the changes, I was able to successfully upload picture with the new patient.

kevinzurek:
The pictures seem to be fixed, but the diabetes prompt is also being disrupted. It needs to be turned on in the Admin config and then it will prompt everytime you have a high BP / high weight (try BP 200/200, height 6'3", and weight 300 with the option turned on - can't remember the exact threshold).

I think the logic here should be along the lines of:
1) User clicks submit
2) If diabetes prompt is not triggered, show the duplicate alert. Done.
3) If diabetes prompt is triggered, do not show the duplicate alert
4) User clicks Yes/No on diabetes prompt
5) Show the duplicate alert. Done.

Once the diabetes prompt is triggered, this line stops the traditional submit by returning false:
https://github.com/FEMR/femr/pull/550/files#diff-08ff3973c54207c832f9ae34bff8753cR628

I think "pass" could control when to show the duplicate alert, or it might be cleaner to create 2 if statements. 1 to reflect steps (1) and (2) above and another to reflect (1), (3), (4), and (5). 

if pass equals true and diabetes prompt is necessary:
-do (1), (3), (4), (5)
else if pass equals true
-do (1), (2)

C-RICHMOND:
@kevinzurek okay, sorry for the mess! I will make my fixes and commit tomorrow. Ill make sure to do thorough testing as well; to avoid any more bugs. Thank you for you help.

kevinzurek:
no worries!! thank you :+1:

C-RICHMOND:
@kevinzurek, just pushed a new commit. Everything seems to be working after the testing I've done, but let me know if you encounter any issues. Sorry for the delay on this!

C-RICHMOND:
@kevinzurek Just realized that I accidentally included a commit from my current change request into this one. It was the creation of a new endpoint to be used for finding and returning matching patients. While it won't be used for this change request, it will definitely be used for the next one, and including it here won't affect any existing code.

kevinzurek:
@C-RICHMOND no problem, thanks for letting us know. Which commit is it?

C-RICHMOND:
@kevinzurek It was commited 7 days ago, 	Merge remote-tracking branch 'refs/remotes/origin/subtask-FEMR149-all...' 

id: dc8a41b

jcreigh:
Warning, here be dragons (rebasing is dangerous). Note: I didn't actually verify your code after doing this, but it looks alright.

For future reference, you should be rebasing in order to keep your branch up to date instead of repeatedly merging yours with master.

Anyway, you can change the parent of your most recent commit (b147da6) to something else with a rebase. Making sure your master is up to date with upstream, replaying everything after dc8a41b onto 6e5d635, and then squashing all your merges by rebasing onto master.

````
git checkout master
git pull upstream
git checkout task-FEMR-148-checkPatientAlreadyExists
git rebase --onto 6e5d635 dc8a41b
git rebase master
```

This results in your history being:
```
% git log --graph --decorate --pretty=oneline --abbrev-commit
* 0793358 (HEAD -> task-FEMR-148-checkPatientAlreadyExists) Restructured code to avoid diabetes propt from being disrupted
* dce39e7 Fixed picture posting error
* 041021c Revert "Test"
* f886ddd Alerts user if matching patient exists
* 1697d6a Test
* 8986e4f (origin/master, origin/HEAD, master) [FEMR-241] Export inventory reports as a CSV
[...]
```

But while you're rebasing (if you want the learning experience), you might as well drop the test commits, so use this to start interactive rebasing.

`git rebase -i master`

And then drop the test commits

```
drop c25145c Test
pick 622e0b5 Alerts user if matching patient exists
drop acfbffb Revert "Test"
pick 4d27848 Fixed picture posting error
pick 3297b5b Restructured code to avoid diabetes propt from being disrupted
```

That results in a history like this:
```
% git log --graph --decorate --pretty=oneline --abbrev-commit
* 21bb7ce (HEAD -> task-FEMR-148-checkPatientAlreadyExists) Restructured code to avoid diabetes propt from being disrupted
* 1a45b28 Fixed picture posting error
* 2f0fa95 Alerts user if matching patient exists
* 8986e4f (origin/master, origin/HEAD, master) [FEMR-241] Export inventory reports as a CSV
[...]
```

And now just push your new history, with the -f flag to force it (since the history has been rewritten).

`git push -f origin task-FEMR-148-checkPatientAlreadyExists`

C-RICHMOND:
@jcreigh Yeah, I'm still trying to get the hang of git; slowly, but surely. Thanks for the heads up!

kevinzurek:
@C-RICHMOND excellent, everything works perfectly. 

1) Fixed up some formatting
2) Removed the commit that doesn't belong (thank you @jcreigh , i had never used `git rebase --onto` like that before)
3) Merged manually:
https://github.com/FEMR/femr/commit/2d00ffc98559f380e8da509fa738fc2bc05c09c7

PR #552, opened 2016-11-04
[Task 245] lists triage patients 

mtwins:
Lists triage patients checked in on the current day on a html page.
-Note: doesn't yet take into accounts multiple trips on the same day( trip id's)

https://teamfemr.atlassian.net/browse/FEMR-245


conf/routes
```diff
@@ -29,6 +29,11 @@ POST        /admin/trips/edit/:id                              @femr.ui.controll
 GET         /admin/trips/teams                                 @femr.ui.controllers.admin.TripController.teamsGet()
 POST        /admin/trips/teams                                 @femr.ui.controllers.admin.TripController.teamsPost()
 GET         /admin/trips/cities                                @femr.ui.controllers.admin.TripController.citiesGet()
+#add test (Hello World)
+GET         /admin/triagePatients/triagePaitent                 @femr.ui.controllers.admin.TriagePatientController.triagePatient()
+POST        /admin/triagePatients/triagePaitent                 @femr.ui.controllers.admin.TriagePatientController.triagePatient()
```
kevinzurek: POST is only used when sending a request/data back to the server. I don't think we need this here.

app/femr/business/services/system/EncounterService.java
```diff
@@ -535,4 +536,47 @@ private Integer getTabFieldIdByTabFieldName(List<? extends ITabField> tabFields,
...
+//troubleshooting
+
+    public ServiceResponse<List<PatientEncounterItem>> returnCurrentDayPatientEncounters()
+    {
+        ServiceResponse<List<PatientEncounterItem>> response = new ServiceResponse<>();
+        List<PatientEncounterItem> patientEncounterItems = new ArrayList<>();
+        //gets all patients encounters
+        ExpressionList<PatientEncounter> query = QueryProvider.getPatientEncounterQuery()
+              .where()
+                .ge("id",0);
```
kevinzurek: It looks like you are going to the database to grab ALL patient encounters then looping over each one to check if it is equal to the current date. In a real system this would pull just over 5,500 encounters  today which shouldn't be an issue, but performance will be impacted as the number of encounters grow

You should be able to use the same logic on dates to reduce the number of encounters being fetched from the database. Instead of .ge("id",0), you should be able to do something like: 

.where()
.ge("date_of_triage_visit", 12:00:00am today)
.le("date_of_triage_visit", 12:00:00am tomorrow)

and you'll automatically get back the encounters you are looking for without the for loop

kevinzurek:
Hi @mtwins , looks like there's a lot of extra changes submitted (comments, commas, spaces, etc... ). I also made a couple comments. thank you!! :+1:

mtwins:
Hi Kevin ,

I will make those changes as soon as possible. Also, my license for  IntelliJ IDEA has run out, so can I get the license key?

Thanks,

Melissa Heredia

mtwins:
Hi,

I have made the changes requested. Let me know if there are any issues.

Thanks!

Melissa

app/femr/business/services/system/EncounterService.java
```diff
@@ -535,4 +535,38 @@ private Integer getTabFieldIdByTabFieldName(List<? extends ITabField> tabFields,
 
         return tabFieldId;
     }
+
```
kevinzurek: While not technically needed, it's good practice to add @Override annotation above this method to inform the compiler that you are overriding the method from IEncounterService. Same with the {@inheritdoc}. I noticed there's other methods that are missing this as well so i will edit before merging.

kevinzurek:
@mtwins no technical issues and very clean :)

Can we isolate this work in a section called "Manager" so it doesn't overlap with Triage? In other words:

app/femr/ui/controllers/admin/TriagePatientController.java becomes
app/femr/ui/controllers/admin/ManagerController.java

app/femr/ui/models/triage/ManageViewModelPost.java becomes
app/femr/ui/models/admin/manager/ManageViewModelPost.java

etc...

This will be its own module, so we want as little overlap as possible with the concept "Triage" in the code

mtwins:
Hi I have refactored the code in a manager section.    Thanks,

Melissa

kevinzurek:
Did some formatting and merged manually, thank you @mtwins  :+1: 

Also, added @Override annotation for the new service method

PR #556, opened 2016-11-22
[FEMR-249] Add script to retain diabetes prompt data during upgrade f…

kevinzurek:
Found columns being destroyed during a 2.1.8 upgrade 

https://teamfemr.atlassian.net/browse/FEMR-249


kevinzurek:
Need immediately, merging manually

https://teamfemr.atlassian.net/browse/FEMR-249

PR #557, opened 2016-11-22
[FEMR-251] Add script to retain diabetes prompt data during upgrade f…

kevinzurek:
Evolutions were merged into existing evolutions which require a full execution of a down after being changed. This one in particular was evolution 98.

https://teamfemr.atlassian.net/browse/FEMR-251


kevinzurek:
Need immediately, merged manually

PR #560, opened 2016-11-28
feature-FEMR157-OptionalPhoneNumberRecord-TakeTwo

Thraxus:
Added phone number record to Triage and SQL.  Also made it appear on the search results and history pages.  The next change will allow patient searches by phone number.

Note that the SQL check on phone number entry was removed.  It was kept strictly client side and handled via java script.  Without intentionally trying to spoof data, the phone number field should be restricted to numeric entries only.

https://teamfemr.atlassian.net/browse/FEMR-157


app/femr/data/models/mysql/Patient.java
```diff
@@ -38,6 +38,8 @@
     private String firstName;
     @Column(name = "last_name", nullable = false)
     private String lastName;
+    @Column(name = "phone_number", nullable = false)
```
kevinzurek: Why are we trying to enforce `nullable = false`? I noticed the empty string gets mapped in the datamodelmapper instead of NULL being used

Thraxus:
Because my fingers are large and in charge, and weren't paying attention to my brain whilst typing.  Fixed now though, good catch.

app/femr/common/ItemModelMapper.java
```diff
@@ -187,11 +189,14 @@ public PatientItem createPatientItem(int id,
         patientItem.setId(id);
         patientItem.setFirstName(firstName);
         patientItem.setLastName(lastName);
+        patientItem.setPhoneNumber(phoneNumber);
```
kevinzurek: you aren't

app/femr/data/DataModelMapper.java
```diff
@@ -363,6 +363,7 @@ public IPatient createPatient(int userID, String firstName, String lastName, Dat
         patient.setUserId(userID);
         patient.setFirstName(firstName);
         patient.setLastName(lastName);
+        patient.setPhoneNumber(phoneNumber);
```
kevinzurek: needs a NULL check otherwise the phone_number column will always be an empty string instead of NULL

Thraxus: Phone_number currently shows null for me in the DB when no phone number is entered.  

The column generates at default and has default = null, so when it receives an empty string it automagically assumes null.

ALTER TABLE patients ADD COLUMN phone_number VARCHAR(20) NULL DEFAULT NULL AFTER last_name;

kevinzurek: let me try fetching and pulling down again, 1 sec

kevinzurek: still getting it (the first 2 results are an empty string from not entering anything):

mysql> select p.phone_number, pe.date_of_triage_visit FROM patients p INNER JOIN patient_encounters pe ON pe.patient_id = p.id WHERE p.phone_number IS NOT NULL; 
+--------------+----------------------+
| phone_number | date_of_triage_visit |
+--------------+----------------------+
|              | 2016-11-30 17:41:00  |
|              | 2016-11-30 21:25:10  |
| 55522278989  | 2016-11-30 21:30:20  |
+--------------+----------------------+
3 rows in set (0.01 sec)

app/femr/common/ItemModelMapper.java
Thraxus: Yeah, this one was a mistake.  Problem still exists though.

app/femr/ui/models/triage/IndexViewModelPost.java
```diff
@@ -77,14 +78,20 @@ public void setFirstName(String firstName) {
         this.firstName = firstName;
     }
 
-    public String getLastName() {
-        return lastName;
-    }
+    public String getLastName() { return lastName; }
```
jcreigh: Did you mean to commit this style change? It makes it inconsistent with the others and seems unrelated to the task.

app/femr/common/ItemModelMapper.java
```diff
@@ -18,6 +18,7 @@
 */
 package femr.common;
 
+import com.avaje.ebeaninternal.server.lib.util.Str;
```
Thraxus: Removing this import for the next commit.  It's unnecessary.

Thraxus:
Should all be fixed now.  Not sure why NULL was working without the inject protection before, but it's working now.

Fixed a few other minor issues as well, and included the history / encounter page in the phone number output.

app/femr/ui/views/history/indexEncounter.scala.html
```diff
@@ -51,6 +51,7 @@
                         <p><span class="infoLabel">First Name:</span> @outputStringOrNA(patient.getFirstName)</p>
                         <p><span class="infoLabel">Last Name:</span> @outputStringOrNA(patient.getLastName)</p>
                         <p><span class="infoLabel">Address:</span> @outputStringOrNA(patient.getAddress)</p>
+                        <p><span class="infoLabel">Phone Number:</span> @outputStringOrNA(patient.getPhoneNumber)</p>
```
Thraxus: I had to break the formatting I've been using on this change with this phone number output.  The style I've been sticking to is: First Name, Last Name, Phone Number.  However, formatting of the page wouldn't allow this to display properly, so I pushed it down under Address.  If this is a problem, I can change it back.

jcreigh:
Not sure why it'd be working before either. NULL is a special value. It's not the same as an explicit empty string. You were setting it to an empty string, so it was being set to an empty string.

> The column generates at null and has default = null, so when it receives an empty string it automagically assumes null is what should stay there

Thraxus:
The only thing I can think of is somehow the ItemModelMapper was protecting the empty string injection.  I put the same protection into the DataModelMapper and now it's working as intended.  

If you look back at #546 though, I never had that protection in place and it worked just fine - or I'm just remembering the past in a way I want to.  Regardless, I don't remember seeing a blank in the DB before Kevin pointing it out (and the screen shot above shows it working too).

kevinzurek:
It's always best to check the string for empty or null so you can control the values. Empty string can have very different behaviors than NULL. As for the old pull request i am not sure how NULL would get into the database, but i would guess 1) if you did a down evolution then an up evolution for this then you would have rebuilt the column with NULL as the default value or 2) your computer has become self aware

either way, this seems to work so i'll merge :+1:

kevinzurek:
merged manually

PR #562, opened 2016-12-05
Improvement femr 126 add descriptions for each setting

judyvbaxter:
Descriptions populate for each system setting and display on the Configure page. 

https://teamfemr.atlassian.net/browse/FEMR-126

### DIFF
--- a/app/femr/ui/models/admin/configure/IndexViewModelGet.java
+++ b/app/femr/ui/models/admin/configure/IndexViewModelGet.java
@@ -4,6 +4,7 @@ import java.util.HashMap;
 import java.util.Map;
 
 public class IndexViewModelGet {
+    //First map contains Name and isActive pair, second contains Name and description pair
     private Map<String, Boolean> settings;;
     private Map<String, String> descriptions;;
 
--- a/app/femr/ui/models/admin/configure/IndexViewModelGet.java
+++ b/app/femr/ui/models/admin/configure/IndexViewModelGet.java
@@ -4,21 +4,32 @@ import java.util.HashMap;
 import java.util.Map;
 
 public class IndexViewModelGet {
-    private Map<String, Boolean> settings;
+    private Map<String, Boolean> settings;;
+    private Map<String, String> descriptions;;
 
     public IndexViewModelGet(){
         this.settings = new HashMap<>();
+        this.descriptions = new HashMap<>();
     }
 
     public Map<String, Boolean> getSettings() {
         return settings;
     }
+    public Map<String, String> getDescriptions() {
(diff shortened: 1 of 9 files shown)


app/femr/ui/models/admin/configure/IndexViewModelGet.java
```diff
@@ -4,21 +4,33 @@
 import java.util.Map;
 
 public class IndexViewModelGet {
-    private Map<String, Boolean> settings;
+    private Map<String, Boolean> settings;;
+    private Map<String, String> descriptions;;
```
kevinzurek: @judyvbaxter can you comment what the Maps contain so someone new can understand why we have 2 maps here? :+1:

judyvbaxter: Done, pushed up the changes. Added a comment that clarifies that the first map is a name-isActive pair, and the second is a name-description pair.

app/femr/ui/models/admin/configure/IndexViewModelPost.java
```diff
@@ -4,12 +4,21 @@
...
+        return descriptions;
+    }
+
     public void setSettings(List<String> settings) {
         this.settings = settings;
     }
+
+    public void setDescriptions(List<String> settings) {
+        this.descriptions = descriptions;
+    }
```
kevinzurek: Do we need these methods in the POST view model?

app/femr/ui/models/admin/configure/IndexViewModelPost.java
```diff
@@ -4,12 +4,14 @@
...
 public class IndexViewModelPost {
     private List<String> settings;
+    private List<String> descriptions;
 
     public List<String> getSettings() {
         return settings;
     }
 
-    public void setSettings(List<String> settings) {
-        this.settings = settings;
```
kevinzurek: @judyvbaxter removing this causes an error when you submit the settings. This is supposed to catch the settings during a POST

judyvbaxter: I've added this method back in.

kevinzurek:
I don't think this is needed either in the POST view model:

https://github.com/FEMR/femr/pull/562/files#diff-e45bd89c919bfb9100b1e46517054458R13

the POST view model should only contain variables for data you are trying to capture from the web page when a user clicks submit and sends the data back to the server. The descriptions aren't necessary to retrieve (they are just there for the user to see).

kevinzurek:
Fixed formatting, rebased, and merged manually

PR #559, opened 2016-11-27
Feature 244 moniter clinic patient flow

mtwins:
Created new tab for manager(Only users with manager status can access/see tab). The tab displays a table of patients(checked in on the current date on the trip the manager is assigned to). The table has columns for patient ID, name, Other patient information, chief complaint, patient vitals, hpi narrative, and check in times for all the 3 stations, as well as a turn around time column which calculates the turn around time. Only patients with open encounters are displayed.

PR #563, opened 2016-12-05
Feature 244 final manager tab

mtwins:
Second Attempt at pull request.
Created new tab for manager(Only users with manager status can access/see tab). The tab displays a table of patients(checked in on the current date on the trip the manager is assigned to). The table has columns for patient ID, name, Other patient information, chief complaint, patient vitals, hpi narrative, and check in times for all the 3 stations, as well as a turn around time column which calculates the turn around time. Only patients with open encounters are displayed.


kevinzurek:
Hi @mtwins can we close #559 ?

I pulled this down and created a new patient, but they aren't being displayed. I pulled the date of triage visit for the new patient:

`select date_of_triage_visit from patient_encounters where patient_id = 4667;`
+----------------------+
| date_of_triage_visit |
+----------------------+
| 2016-12-06 20:11:14  |
+----------------------+
1 row in set (0.00 sec)


Here's the manager screen afterwards:

manager

app/femr/ui/models/manager/IndexViewModelGet.java
```diff
@@ -33,5 +45,108 @@ public String getToday()
...
+        return encounterItem.get(i);
+    }
+
+    // gets array list of Patient Encounter Items
+    public List<PatientEncounterItem> getEncounter() {
+        return encounterItem;
+    }
+
+//returns turn around time in the form of a String
+    public String turnAroundTime(PatientEncounterItem item){
```
kevinzurek: A few suggestions for this viewmodel that you build - would like to hear your thoughts as well. Viewmodels are supposed to be "dumb" in that they don't do much logic. This `turnAroundTime()` method may come in useful elsewhere. i would consider moving the logic to our utilities:

https://github.com/FEMR/femr/tree/master/app/femr/util/calculations

and invoking it when needed (should be in the controller)

app/femr/ui/models/manager/IndexViewModelGet.java
```diff
@@ -33,5 +45,108 @@ public String getToday()
...
+        String triageTime;
+        if(item.getTriageDateOfVisit()!=null) {
+            String[] parts = item.getTriageDateOfVisit().split("-");
+            triageTime = parts[1];
+            return triageTime;
+        }
+        else
+            return " ";
+    }
+    public String getMedicalTime(PatientEncounterItem item) {
```
kevinzurek: It looks like `getPharmTime()`, `getTriageTime()`, and `getMedicalTime()` all do the same exact thing with a string that is formatted in a specific way as input. Could these all be combined into one method that accepts the string as a parameter (it doesn't matter if it's triageDateOfVisit, medicalDateOfVisit, etc...) . I would consider moving this into a utility as well. 

The ViewModel should really only have a variable called triageTime, medicalTime, etc... that gets called by the view.

mtwins: image

Hmm that's weird, it is working on my page.

kevinzurek: looking a bit deeper, only patients with more than one chief complaint are being displayed because of the conditional if statement that this is nested in:
https://github.com/FEMR/femr/pull/563/files#diff-559b23d6ab77babeb5d791a98ed6d4c6R96

It is hard to  remember what "p", "v", and "i" all stand for when reading the code as someone who didn't write it :) . while technically correct, it's probably worth making the variable names meaningful

mtwins:
I pushed some more changes for the manager tab(Streamlined the view model and moved the functions to date utils).  I am planning to simplify the html page more this weekend.

kevinzurek:
Looking much better. Think of the IndexViewModelGet as a dumb class that contains data for the view. It should mostly be getters and setters. Instead of doing `return getDisplayTime(item.getTriageDateOfVisit());`, just store the string itself in the view model (set it by calling the new utility in the controller).

In the controller:
`viewmodel.setDateOfTriageVisit(getDisplayTime(item.getTriageDateOfVisit()))`
In the view:
`viewmodel.getDateOfTriageVisit()`

Now that you have most of the functionality working, take a look back at what you wrote and add comments that you think will help people in the future and clean up the formatting :). little things like the indentation on this return statement will add up to look really messy:

https://github.com/FEMR/femr/pull/563/files#diff-14a7b59e7e664ca20db50af087bf4231R57

In IntelliJ you should be able to auto format by highlighting the code and pressing ctrl+shift+alt+L (default hotkey).

mtwins:
I pushed another revision with the updated viewModel.

mtwins:
Hi, I noticed an error in vital display on table, I will update the code .

mtwins:
Fixed the errors with displaying information on the html table and cleaned up the display and updated the header ect.

kevinzurek:
Refactored a bit and moved to new pull request:

https://github.com/FEMR/femr/pull/587

PR #564, opened 2016-12-12
Subtask femr149 allow user confirm patient match

C-RICHMOND:
Added functionality to alert user if there are matching patients in the database. The functionality includes: new endpoint to return list of matching patients, new service to update existing patient information, modal window to display matching patient information, and various js changes to account for post submission.


C-RICHMOND:
NOTE: Up until now I have been using the github desktop app for commits, branching, etc. This was a horrible mistake, and will be using git from here on out. With that in mind, looking at my commits for this pull request, I'm not sure why all the "merge remote-tracking branch" commits were included. If you need me to try and clean this up I can; I just need some direction. Otherwise, let me know if I need to make any changes to the code. There were a little of different parts in this change request, and it is possible I may have missed something.

C-RICHMOND:
One more thing, when an existing patient is updated with new information, the user is taken to the patient history page as normal. On this view page, the user has the option to view previous patient encounters. If a user clicks on a previous encounter, all the vital information is accurate for that specific encounter (i.e. ); however, the patient information is not completely accurate; it has been updated. For example, say a patient just went through triage and they were an existing patient. It turns out they had come through a year ago as well, and since then they are at a different address, different age. If the user clicks on previous encounters, they will see the previous vitals, but the patient information from a year ago will not display; instead the new address, new age, and new picture will display (it wont show the patient information from a year ago). Is this okay?

kevinzurek:
The commit history might be ugly, but the diff is nice and clean :+1: . It's not necessary to clean up the commit history because it will be squashed before merging. If you wanted to cleanup you could probably do so with a `git rebase master`. Just be careful as this rewrites your commit history. The commit history looks like this because i'm assuming, via the desktop app, you are merging updates from the master branch into your feature branch.

Your second question is a good one. Displaying the most recent data on the page isn't an issue, but I think the issue becomes can we just overwrite the old data? It looks like the `updatePatient(id, patientItem)` method will overwrite any updated attribute you give it now that it can accept the entire patientItem as a parameter. I have 2 concerns:
1) Can we overwrite that data?
2) Is it good practice to write a generic `updatePatient()` method?

Right now my thoughts are:
1) No because as developers we don't have an excuse to "hard delete" data. None of us can foresee doctors or researchers needing / not needing this data and it's technically feasible to retain it.
2) It works right now... Once you start introducing anymore logic or if you try to unit test this method you'll run into issues. Eventually this could be broken out into `updateFirstName()`, `updateLastName()`, `updateCity()`, etc...

C-RICHMOND:
I understand your concerns. Speaking to that, if there is a matching patient found, but they have aged since their last visit or changed address, shouldn't we update that information? The only information that is potentially changed by updatePatient() is the basic information: name, city, address, birthdate, sex. The vitals are not touched, and vitals from previous encounters display on the 'previous encounter' page.

C-RICHMOND:
Also, while I am accepting the entire patientItem as a parameter, I am not using all the member variables. Like I mentioned above, I am only using the basic patient information (see comment above for specifics) in the service function. It was just easier to pass the patientItem instead of the six individual variables.

kevinzurek:
I think the `updatePatient()` method is okay for now. The reason for this concern in the long run is that the patientItem object can easily change externally and impact the method. We have select other methods that have this happening as well, it's additional complexity to manage inside `updatePatient()`. I don't have a lot of experience unit testing, but unit tests also seem to become more difficult due to different paths through the same method.

Yes, we should absolutely be able to update the information for everything you've updated in `updatePatient()`. It becomes a question of whether or not we track the updates - 
Consider 1 example:
1) Dr. 1 sees a new patient "Bob Smith" from Detroit
2) A year later "Bob Smith" comes back and sees Dr. 2 who finds out the first name is not bob and that he's moved, "Bob Smith" is now "John Smith" from Chicago
3) Dr. 1 tries to look up "Bob Smith" 

fEMR will have no memory of Bob moving or changing his name. I'll get some feedback from our users about this later today.

C-RICHMOND:
Okay, yeah I see your point. Just let me know what direction you would like me to take, and I can modify the logic accordingly.

kevinzurek:
A resounding yes for keeping the data. 

Lets create a new JIRA ticket for being able to store historical general patient info as it wasn't a requirement for FEMR149. It can be a dependency for this pull request in its current state

C-RICHMOND:
Okay, so in terms of this pull request does anything need to be changed? I can work on that dependency next.

kevinzurek:
No, lets keep this as-is.

kevinzurek:
Created FEMR-265:

https://teamfemr.atlassian.net/browse/FEMR-265

PR #565, opened 2016-12-17
FEMR-185 : Change the name of the "Treatment Given" field

axm155231:
Change the name of the "Treatment Given" field to "Procedure/Counseling"
Change the name of this field in the tab_fields table to reflect the new name.


kevinzurek:
Hi @axm155231 , a global search for the concept "Treatment" needs to be done in the project. I think there's a few other places that use this as it is named. I think that, for example, the below line would break after the renaming:

https://github.com/FEMR/femr/blob/d929442f52b84b95c6214028d893db33ce516e35/app/femr/ui/controllers/MedicalController.java#L209

axm155231:
As I remember, the change request was about changing the  ‘treatment given’ field in the ‘tab_fields’ table, hence
we did not change the tab name itself which I vaguely remember comes from the ‘tabs’ table.

When we were working on this, we saw

i)                    Field named ‘treatment’ in the tab_field table (which we change to procedure_counseling)

ii)                   Filed named ‘Treatment’ in tabs table (which left as it is)

We also inspected the DatabaseSeeder.java file and we felt confident that these are two separate things.

In our local testing we did not encounter breaks with leaving the tab name as it is

kevinzurek:
You are correct about the tab, @axm155231 ! I did a find all and found 2 more places i think it should be changed. Please see this commit:

https://github.com/kevinzurek/femr/commit/3229ac99d90019ee7aa88cfb7110df360e023609

If you agree with this I will merge. Thank you!!!

axm155231:
I do not see any issue with that.

kevinzurek:
Merged manually and fixed duplicate ticket #s in JIRA.

master: 284de9780b5d002095ec910b804072232ef22cb8

Thank you! :)

PR #566, opened 2016-12-18
Improvement.FEMR208: pdf not displaying amount of prescription dispensed

ShaneQi:


kevinzurek:
Hi @ShaneQi , i merged the code from @gerrytucker78 since it covered the form displaying as "N/A" instead of "NULL". thank you guys!!

https://github.com/FEMR/femr/pull/567

PR #567, opened 2016-12-18
Improvement 208 encounter pdf display amount dispensed

gerrytucker78:
Workaround in place for potentially new issue in MedicationRepository when creating a new prescription.

Updated display to match layout of PDF.


kevinzurek:
manually merged into master:

master: d929442f52b84b95c6214028d893db33ce516e35

Removed Todo comment for now, good catch. thanks @gerrytucker78 !

PR #568, opened 2016-12-18
femr-137: Added ageCalculated field to patients DB table, displays, services

gerrytucker78:
While testing, noticed an issue exists with entering an invalid date for birthday that existed prior to changes.  Did not attempt to resolve this issue with the new feature.

PR #570, opened 2016-12-21
Improvement.FEMR137: add flag for fake birthday

ShaneQi:


app/femr/ui/views/triage/index.scala.html
```diff
@@ -63,6 +63,7 @@ <h2 class="text-center">Check In - Triage</h2>
                     @inputAge("Age", "Years", "years", "Months", "months", if(viewModel != null) viewModel.getPatient else null)
                     <span class="orSpan">OR</span>
                     @inputDate("Birth Date", "age", if(viewModel != null) viewModel.getPatient.getBirth else null)
+                    <label class="checkBoxLabel"><input type="checkbox" name="birthdayIsFake"> BIRTHDAY IS FAKE</label>
```
kevinzurek: The system should be able to identify if the birthday is fake behind the scenes (this could be a hidden checkbox that gets checked by JavaScript when a birthdate is generated automatically). something similar to what @gerrytucker78 did in the javascript file here:
https://github.com/FEMR/femr/pull/568/files#diff-08ff3973c54207c832f9ae34bff8753cR590


creating extra clicks for a user should only be done if absolutely necessary!

kevinzurek:
closing pending fixes

PR #574, opened 2017-01-11
FEMR-217 Fix Medical tab

CEN5035-4021-FSU:
- Updated patientOverview.scala.html view to reflect changes to pageName argument passed by medical/edit.  As of rev 23c533c, edit no longer passes pageName "Medical", but instead passes "Medical - "...  Update changes condition related to BMI display to check if argument passed to pageName contains "Medical -".

Contributed by Jonathan Clow during the CEN5035 course at FSU


kevinzurek:
Merged manually & added a comment above the patientOverview() call to stop this from happening again.

https://teamfemr.atlassian.net/browse/FEMR-217

master: 17411f66a97d7fd082ca057e84c65047108d4c70

PR #576, opened 2017-01-11
[FEMR-113] Added check for null height in PDF to fix bug

CEN5035-4021-FSU:
If a patient's height is not provided, it will be displayed as N/A in the
generated PDF of an encounter. Previously, a null height was displayed as 0'0", which
could easily cause confusion.

PDFController now includes value checking for both heights in feet and inches.

Contributed by Robert Massicotte during the CEN5035 course at FSU

https://teamfemr.atlassian.net/browse/FEMR-275


kevinzurek:
Merged manually & created FEMR-275 to take care of the same issue in the encounter history view

master: 0fd7ab6f41708cafb8d937eac3d907ec7f0a6356

PR #577, opened 2017-01-11
[FEMR-151] Add age group and age value validation in triage (frontend)

CEN5035-4021-FSU:
Implements error checking on client side to detect a mismatch
between the entered age and age group during triage.

Years and months are used to compute age, and this age is then compared with the given thresholds.
There is an small validation that checks when the user does not provide years or months (so it assigns a default value of 0)

The form cannot be submitted until the inconsistency is solved.

Contributed by Jarrod Moore during the CEN5035 course at FSU

https://teamfemr.atlassian.net/browse/FEMR-151


kevinzurek:
Excellent! Merged manually

master: e99a36d52ad6239ad89e0812ec36e1815ad4ca7f

PR #579, opened 2017-01-12
[FEMR-208] Display prescription amount in encounter's PDF

CEN5035-4021-FSU:
Get the prescription amount value and display it in the dispensed
prescription part of the generated PDF.

PDFController now adds a new column to the report with the amount dispensed.

//TODO: Please verify the behavior of this change with replaced medications. Not sure if the current output is the expected one.

 Contributed by Xiaoxiao Gan during the CEN5035 course at FSU


kevinzurek:
Fixed in https://github.com/FEMR/femr/pull/567

PR #588, opened 2017-03-19
Bug femr304 typeahead broken

kevinzurek:
Fix to stop patients from being saved with no value in city and a script to fix a few records that made it in previously

https://teamfemr.atlassian.net/browse/FEMR-304


kevinzurek:
Merged manually

master: 035faf901d0c4848b71af1a25ed84e8b85469b24

PR #589, opened 2017-03-22
redirect logging from stdout to logging file

kevinzurek:
https://teamfemr.atlassian.net/browse/FEMR-305

Adds an optional logger to redirect logging from stdout to a file. To use, set the environment variable 'logger.file' to target conf/prod-logger.xml.


kevinzurek:
https://teamfemr.atlassian.net/browse/FEMR-305

PR #592, opened 2017-03-26
Adding validation based on ticket 278

SamuelDev:
JS and model validation both added for ticket 278.

https://teamfemr.atlassian.net/browse/FEMR-278


kevinzurek:
Merged: 6f7d9127a4714dbeebfe344688f3ea4e20ca2498


This is sufficient for FEMR278 as it's written. This interface is still confusing as hell for a user. Adding FEMR-306 to cover below scenarios:

1) Clicking submit without filling any fields results in an error
2) Filling out only Brand/Quantity results in an error

kevinzurek:
Ran into issues adding a medication from the concept dictionary based on the validation logic. Applied a quick fix for 2.3.3 release:

0d59386509faba0794b4c4af183bb8f4f1455aad

PR #593, opened 2017-03-27
JIRA Bug 232 - Fix for display issue in encounter history where replacement prescription displays incorrect amount

drakesvoboda:
…cement

prescriptions dislayed incorrect amounts

### DIFF
--- a/app/femr/business/services/system/SearchService.java
+++ b/app/femr/business/services/system/SearchService.java
@@ -419,10 +419,11 @@ public class SearchService implements ISearchService {
                     .collect(Collectors.toList());
             List<PrescriptionItem> replacedPrescriptions = patientPrescriptions.stream()
                     .filter(pp -> pp.getPatientPrescriptionReplacements().size() > 0)
-                    .map(pp -> itemModelMapper.createPrescriptionItem(
+                    .map(pp -> itemModelMapper.createPrescriptionItemWithReplacement(
                             pp.getId(),
-                            pp.getPatientPrescriptionReplacements().get(0).getReplacementPrescription().getMedication().getName(),
                             pp.getMedication().getName(),
+                            pp.getPatientPrescriptionReplacements().get(0).getReplacementPrescription().getMedication().getName(),
+                            pp.getPatientPrescriptionReplacements().get(0).getReplacementPrescription().getAmount(),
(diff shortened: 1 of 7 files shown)


kevinzurek:
Hi @drakesvoboda - this is a good use of the PrescriptionItem! I tried a simple test and it looks like the original medication is being printed twice (see below - i replaced Prescription 1 with Prescription 2):
test1

kevinzurek:
Pending 
https://teamfemr.atlassian.net/browse/FEMR-210
https://teamfemr.atlassian.net/browse/FEMR-242

drakesvoboda:
@kevinzurek Take a look at my most recent commit; I completely changed how the prescriptions were displayed on the encounter history page. I also made some minor style tweaks to this page.  a look at the generated pdfs; I made some edits to those also, however they still have the old table format. Let me know what you think.

kevinzurek:
@drakesvoboda Excellent work & thank you for taking initiative to rebuild the prescriptions! They have been buggy for a long while now :+1: 

I think that, to merge this, only 1 change needs to be made:
1) The "original" column in the PDF becomes a bit misleading because the replaced medication still shows up in it. I think it would be better to eliminate the "Replacement" prescription in the "Original" column (see pic below).
pdf

Additional feedback:
A prescription string is made up of many different elements (medication name, generic ingredients, brand names, quantity, form, etc, etc... ). The prescription strings are being built in multiple places: PDFController.java, indexEncounter.scala.html. In reality this string should be the same regardless of where it's being displayed and regardless of if the prescription is an original or a replacement. Making one utility that can be called for a string in the PrescriptionItem is worth considering - similarly to how mission trip titles are generated here:
https://github.com/FEMR/femr/blob/6f7d9127a4714dbeebfe344688f3ea4e20ca2498/app/femr/util/stringhelpers/StringUtils.java#L207

-Kevin

kevinzurek:
Pulling this in for the 2.3.3 release anyway because it looks way better than what we have now :+1: 

Merged manually into master: d19cc58967201b6295e48bdc82e3a4cdf1e9313c

Closes issues:
https://teamfemr.atlassian.net/browse/FEMR-232
https://teamfemr.atlassian.net/browse/FEMR-210
https://teamfemr.atlassian.net/browse/FEMR-242

PR #594, opened 2017-04-04
[FEMR-263] Duplicate patient notification showing on new encounters

adriankalinowski:
Added patient Id as hidden element in triage fields. Patient Id is used in checkIfDuplicatePatient() function to see whether a patient is returning or not. Fixes the issue of duplicate warning message appearing when the patient is only returning.


kevinzurek:
Merged manually into master: 59f48352459bfaec10d84a9b129573ab3f0f25f8

PR #596, opened 2017-04-10
[FEMR-287] Let a user delete a patient encounter

adriankalinowski:
Implemented a Delete Encounter Button in the encounter history page. Once an administrator or superuser clicks on an encounter, they will have the option to delete the encounter in the top right corner. This is a soft delete which records the time, date, user id, and reason for the delete in columns that were created in the `patient_encounters` database. This pull request includes the commits from pull request #594 as well.


kevinzurek:
@adriankalinowski this looks really really good. One question - is it necessary for the TriageController and EncounterService to require a patientId to delete the encounter?

PR #598, opened 2017-04-23
[FEMR 308] Added validation in pharmacy edit and medical edit to ensure that the prescription amount field is not left blank

drakesvoboda:
0 seemed to be the best way to treat a null value for an amount throughout the application. I imagine that trying to handle a Null value would overly complicate the medication inventory. I think this is a pretty good solution.

### DIFF
--- a/public/js/medical/medicalClientValidation.js
+++ b/public/js/medical/medicalClientValidation.js
@@ -1,37 +1,45 @@
 
 var medicalFieldValidator = {
     isValid : true,
+    errors: [],
     validatePrescriptions : function(){
         var prescriptions = [];
         prescriptionFeature.refreshSelectors();
         $(prescriptionFeature.allPrescriptions).each(function(){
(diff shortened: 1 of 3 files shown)


kevinzurek:
@drakesvoboda I think normally you'd be correct here, but due to the nature of these trips we need to restrict as few fields as possible. This is especially true on the medical page.

What complications would be introduced?

drakesvoboda:
Ok, so I went back and allowed the amount to be null and removed the added validation. My only issue was that in order to complete this change, I needed my changes from the [FEMR 309] pull request. I merged the two branches into a third branch (I'm not sure if this is what I was supposed to do) then pushed that to my fork. Should I now submit a pull request for the third branch? It includes my changes for both 308 and 309.

kevinzurek:
Yeah if you have a 3rd branch with both 308 and 309 in it just go ahead and do 1 pull request with that branch :)

kevinzurek:
Merged in #600

PR #599, opened 2017-04-23
[FEMR 309] Changed how prescriptions are displayed in encounter PDF

drakesvoboda:
Here is a sample PDF with the updated prescriptions table.

[SampleEncounter.pdf](https://github.com/FEMR/femr/files/950031/SampleEncounter.pdf)


kevinzurek:
Merged in #600

PR #603, opened 2017-06-30
FEMR-295 Error message now appears after failed login

jmflorezff:
only the template and three create.render calls affected.


kevinzurek:
Simple and to the point :)

Merged into master: 93f1fed814d6b30b7f29816d74b0d395522bc3f8

https://teamfemr.atlassian.net/browse/FEMR-295

PR #607, opened 2017-06-30
FEMR 276 BF Change Duplicate Record To Show Up When Appropriate

jmflorezff:
Added the city to the query sent to doesPatientExist. Made
doesPatientExist function trigger the display of duplicate record dialog
box only when patient being submitted does not completely match current
records and the first and last names match other existing record(s).

### DIFF
--- a/app/femr/ui/controllers/SearchController.java
+++ b/app/femr/ui/controllers/SearchController.java
@@ -96,12 +96,28 @@ public class SearchController extends Controller {
      * @return
      */
     public Result doesPatientExist(String query){
-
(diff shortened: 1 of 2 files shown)


jmflorezff:
I am not sure the behavior is completely correct, but let me know and we can work on it.

kevinzurek:
@jmflorezff @Cerbes going to close this, lets re-open if you are interested in refactoring. The unwanted behavior i'm finding is:

1) Create a new patient
2) Create another new patient with the same first name, last name, age, and city as in step (1)
3) The duplicate alert won't show up

then

4) Create a new patient
5) Create another new patient with the same first name, last name, age, and a different city as in step (4)
6) The duplicate alert will show up

PR #610, opened 2017-11-02
Feature femr 297 inventory time stamp

jwgettel:
https://teamfemr.atlassian.net/browse/FEMR-297


kevinzurek:
@jwgettel i'll wait for you to merge the methods in the ItemModelMapper. 

You can also completely remove the "out/*" folder and commit the deleted files to get rid of the massive number of unnecessary files here. I tried it here and it seems to work fine:
https://github.com/kevinzurek/femr/commit/44545ec49e0840bc53b8980612f3b201b4aa9206

kevinzurek:
I've been poking around and found there's 1 use case that isn't covered and needs to be handled before merging. To reproduce it you need to start with a database that doesn't have the 2 columns & foreign key added yet:

1) Add a medication to the inventory so that one exists prior to applying the database evolutions
2) Apply database evolutions
3) The foreign key constraint evolution will fail

This is because the medication inventory from (1) doesn't have a user id set and we don't know, retroactively, who did it. This means the new columns have to allow a NULL value. I verified this is the issue by allowing NULL values here:
https://github.com/kevinzurek/femr/commit/32ba40268b4bfa52265bd9bd7f58e73bfd3d4f7d

However, this introduces some issues with a accessing null values when an int is expected elsewhere in the code. I didn't get to debugging this part. 


tl;dr: The code needs to take into consideration a NULL value in the `createdBy` column because we don't know (and never will) the user who created inventories in the past

jwgettel:
@kevinzurek Should be good now.

kevinzurek:
@jwgettel awesome, will pull it down this evening 👍

kevinzurek:
Merged into master:

f92fc8723f3a46482ff9142eab55f8181fec6064

PR #611, opened 2017-11-02
Turn blood pressure text red & make it bigger if blood systolic blood…

msaleh97:
… pressure is >=140 diastolic

0


kevinzurek:
@msaleh97 it looks like there's some files missing, there's only an id change on the html element here!

kevinzurek:
replaced with https://github.com/FEMR/femr/pull/613

PR #612, opened 2017-11-03
Fixed an exception that occured when the height was not entered in tr…

osmanghoni1:
…iage, and medical

### DIFF
--- a/app/femr/ui/views/partials/helpers/outputHeightOrNA.scala.html
+++ b/app/femr/ui/views/partials/helpers/outputHeightOrNA.scala.html
@@ -2,16 +2,16 @@
 
 
     @if(isMetric) {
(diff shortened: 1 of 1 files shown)


kevinzurek:
@osmanghoni1 i'm going to hold off on merging this for now because it seems to break the fix implemented by @adamMolner for metric <-> imperial conversions (326bb3b8e200ceb686ca83ca54b053c60f646994)

When i'm on the master branch without this fix and enter a height, this is how it's displayed in medical:
image

After applying the fix in this pull request, the metric value isn't displayed:
image

osmanghoni1:
Okay I will add in @adamMolner's fix to the project and implement a solution that works with it.

kevinzurek:
Merged into master:

ca795b934f3aaf51e4927325f30605117797b0c0

PR #614, opened 2017-11-08
[FEMR-273] Remove Research Only Setting

ajmanor:
Removed "Research Only" option from System Settings.


kevinzurek:
@ajmanor It looks like the Research Only concept comes up again below this change in the same file. 

On a separate note - wouldn't it be better to pull the object from the existing systemSettings list instead of running another query against the database?

kevinzurek:
Took this commit out of #615

PR #616, opened 2017-11-10
Improvement-288

ajmanor:
On Triage/Medical/Pharmacy pages, if a user had information in an field and clicked on a button, they would move to the next page and lose their changes. As a temporary fix, I have made all the buttons on those pages open in a new tab.


kevinzurek:
Merged issue 288 into master:
1be22b4112236127675b32703fc51517b9ce54f9

PR #617, opened 2017-12-01
Feature branch

msaleh97:


conf/evolutions/default/98.sql
```diff
@@ -2,6 +2,12 @@
 
 ALTER TABLE `patient_encounters`
 ADD COLUMN `is_diabetes_screened` BOOL NULL DEFAULT NULL AFTER `mission_trip_id`;
+ALTER TABLE `patient_encounters`
```
kevinzurek: @msaleh97 old evolution files shouldn't be changed - go ahead and add a file 105.sql in this case and put the changes there.

If you change old evolutions the database will downrev to 98.sql and try to rebuild itself

app/femr/ui/controllers/TriageController.java
```diff
@@ -308,6 +308,25 @@ public Result deletePatientPost(int patientId){
         return redirect(routes.TriageController.indexGet());
     }
 
+   public Result deleteEncounterPost(int patientId,int encounterId){
+    //patientEncounter.getId
```
kevinzurek: old comments can go

app/femr/business/services/core/IEncounterService.java
```diff
@@ -74,7 +74,7 @@
 
     /**
      * Creates a bunch of tab fields that belong to a chief complaint
-     *
+     * Method is returning all none deleted encounters
```
kevinzurek: I don't think the comment is necessary on this method because it's only for creating new tab fields. It should be more explicitly commented in methods which are retrieving a patient encounter by id so the developer knows they aren't getting deleted encounters back

conf/routes
```diff
@@ -64,7 +64,7 @@ GET         /triage/:id                                        @femr.ui.controll
 GET         /triage                                            @femr.ui.controllers.TriageController.indexGet()
 POST        /triage                                            @femr.ui.controllers.TriageController.indexPost(id: Integer)
 POST        /triage/:id                                        @femr.ui.controllers.TriageController.deletePatientPost(id: Integer)
-
+POST        /history/patient/:id                              @femr.ui.controllers.TriageController.deleteEncounterPost(id: Integer,ecounterid: Integer)
```
kevinzurek: This doesn't really fit in well with the Triage section, perhaps better in the History section next to the other encounter routes. Since this is an action performed on an encounter and not a patient you will want to use the same template "/history/encounter/action/:id" like the others:

https://github.com/FEMR/femr/blob/master/conf/routes#L48

kevinzurek: This doesn't really fit in well with the Triage section, perhaps better in the History section next to the other encounter routes. Since this is an action performed on an encounter and not a patient you will want to use the same template "/history/encounter/action/:id" like the others:

https://github.com/FEMR/femr/blob/master/conf/routes#L48

conf/evolutions/default/105.sql
```diff
@@ -0,0 +1,22 @@
+# --- !Ups
+
+
+ALTER TABLE `patient_encounters`
+  ADD COLUMN `isDeleted` DATETIME NULL DEFAULT NULL AFTER `date_of_diabetes_screen`;
+ALTER TABLE `patient_encounters`
+  ADD COLUMN `deleted_by_user_id` INT(11) NULL DEFAULT NULL  AFTER `isDeleted`;
```
kevinzurek: user_id needs to have a foreign key constraint to the user's table so the database enforces restrictions on Ids that can be put in this column (they have to exist in the user's table to add them)

You can see an example here:
https://github.com/FEMR/femr/blob/master/conf/evolutions/default/104.sql#L6

test/unit/app/femr/business/services/PatientServiceTest.java
```diff
@@ -46,7 +46,7 @@ public void setUp() throws Exception {
         mockPatientRepository = new MockPatientRepository();
         mockDataModelMapper =  new MockDataModelMapper();
         mockItemModelMapper = new MockItemModelMapper();
-        patientService = new PatientService(mockPatientRepository, mockDataModelMapper, mockItemModelMapper);
+        patientService = new PatientService(mockPatientRepository, patientEncounterRepository, mockDataModelMapper, mockItemModelMapper);
```
kevinzurek: This needs to be initalized before it's used, it throws an error when you run the unit tests and should be red in the IDE.

To run the tests on the CLI with SBT:  `sbt clean compile test`

kevinzurek:
Had to make changes to handle cases where patients with no encounters had to be seen in Triage which led to some other refactoring.  

Merged into master with 2 commits: :+1: 

8f0f18e5bf28d00531d633cc905733ba6e2f3386
3fd6e187d29eaa234a74df0e1b19e542b703282c

PR #619, opened 2017-12-06
[FEMR-284] Add Checkbox for smoking alcohol diabetes (incomplete)

osmanghoni1:
Require some help outputting values to different views.

https://teamfemr.atlassian.net/browse/FEMR-284


osmanghoni1:
I'll be working on this today and will attempt to complete it by the end of the weekend.

kdunlap:
Hey @osmanghoni1 - If I can answer any questions for you send me a PM in fEMR Slack - @kendunlap. I'll be around all weekend.

app/femr/common/ItemModelMapper.java
```diff
@@ -176,8 +176,11 @@ public PatientItem createPatientItem(int id,
...
                                                 String pathToPatientPhoto,
                                                 Integer photoId,
-                                                String ageClassification) {
-
+                                                String ageClassification,
+                                                Integer smoker,
+                                                Integer diabetic,
+                                                Integer alcohol) {
+        // Osman above
         if (StringUtils.isNullOrWhiteSpace(firstName) ||
```
kdunlap: Make sure to actually set the `smoker`, `diabetic`, and `alcohol` values in the `PatientItem` being created here.

conf/evolutions/default/107.sql
```diff
@@ -0,0 +1,30 @@
+# --- !Ups
+
+INSERT INTO `vitals` (`id`, `name`, `data_type`, `unit_of_measurement`)
```
kevinzurek: Since there's no database structure being changed i would propose adding these during startup via a new section in the DatabaseSeeder instead:
https://github.com/FEMR/femr/blob/master/app/femr/util/startup/DatabaseSeeder.java

The reason for this is if you:
1) run the Ups evolution on a system and the vitals are placed into the table
2) a user creates entries in the patient_encounter_vitals table that links to one of the vitals in (1)
3) run the Downs evolutions - the Downs will fail because you will be breaking the foreign key constraint between patient_encounter_vitals and vitals

kevinzurek: Made the changes mentioned above here during testing:

https://github.com/kevinzurek/femr/commit/b0c0bf5b01c7bdd4b42c3d52bfd645a842d78316

kevinzurek:
My only concern before merging this is differentiating between 

1) asking the patient if they are a smoker and the patient saying "No"
and
2) not asking at all

both conditions will result in the same output on the encounter history page and the vitals listed on medical. @kdunlap @osmanghoni1 thoughts?

osmanghoni1:
I suppose I could change it to N/A, this could mean that it was not checked or that they do not. Otherwise maybe a yes, no check box. Not sure what that would take though.

kevinzurek:
Ah yeah that'd be easy - N/A it is!

osmanghoni1:
I think that should do it: https://github.com/FEMR/femr/pull/619/commits/7037d04b38478d7898442afccd1b07b1456510bf

Let me know if there is anything else that needs to be done.

kevinzurek:
merged into master:

bd2b47514ebbcfc1dcd2940c843da35fbb1261e0

@osmanghoni1 awesome work! :+1:

PR #621, opened 2017-12-10
FEMR-317 built on the bug that caused metric conversion to fail. Added the sam…

adamMolner:
…e fix to medical that was applied to triage earlier

This fix that was applied earlier to the Triage Controller was this time applied to the Medical Controller.


kevinzurek:
Excellent catch - I think this is a continuation of FEMR-317:

https://teamfemr.atlassian.net/browse/FEMR-317

kevinzurek:
Merged into master :+1: :

fc606f53b4fcde3d0e22da0212e3697cb8866dc0

PR #622, opened 2017-12-13
Built upon previous push for this bug but this time I made it so that…

adamMolner:
… the encounter values for weight now come from the vitals map

Their was a slight discrepancy in the way that height data was displayed in the medical view when compared to how it was displayed in the encounter view. The problem was due to the fact that data in the medical viewer display was pulled from the vitals map whereas data in the encounter view display was being pulled from the class patientvitals where it was being stored as in int. The encounter view display is now pulling values from the vitals map and displays the correct value for all encounters.

This version of the push made it so the value for weight in encounter is also drawn from the patient vitals map.

### DIFF
--- a/app/femr/ui/views/history/indexEncounter.scala.html
+++ b/app/femr/ui/views/history/indexEncounter.scala.html
@@ -70,12 +70,42 @@
                             </p>
                         }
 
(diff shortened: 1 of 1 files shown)


kevinzurek:
@adamMolner awesome!! thank you -

merged into master:
de89869908fccc53828aef87ce62d6c05fbee8cd

PR #625, opened 2018-01-18
FEMR-274 Updated JQuery to 3.2.1

ajmanor:
JQuery source located in public/js/libraries/jquery-main-min.js and referrenced throughout the app.


kevinzurek:
Saving this for 2.3.8

PR #626, opened 2018-01-21
FEMR-337 - Fixes bug with viewing null quantity prescription encounters

kdunlap:
This fixes an immediate error, but brings up the question of should prescription inventories ever not be an integer? Would a `null` value ever be needed? We can fix the small hole here, but it may be worth considering some more validation or forcing quantity to default to 0.


kevinzurek:
Good catch. I can't envision null ever being needed, but we have had requests for decimals in prescription amounts... I think this is what you mean:

https://teamfemr.atlassian.net/browse/FEMR-290

For this small part i think it would only require changing the `amount` type. it's a good change request for students this semester i think

kevinzurek:
Merged into master:

dc6c439354af2213bd3b7ac3facd0c168f069cb4

PR #627, opened 2018-01-21
FEMR-336 - Fixed problem with creating tab fields on a new encounter

kdunlap:
I can't find any evidence that `isDeleted` was ever added to the `cheif_complaints` table, so I think this code was errantly added in FEMR-287

https://github.com/FEMR/femr/commit/3fd6e187d29eaa234a74df0e1b19e542b703282c#diff-8590efd0690401d52100c4fb74a18040R257


kevinzurek:
I must have overlooked this while accepting the last round of PRs.... thanks for digging it up!

Merged into master:
85e378cb1a7631e3b0b9b2ac7451e5caa133d42c

https://teamfemr.atlassian.net/browse/FEMR-336

PR #630, opened 2018-03-07
Bug 277 fix trip formulary undo

anakhleh:
Fix for fEMR-277. Bug Description: If a medication was from the set of existing medications added to trip formulary, and user hit remove button (the button now reads undo) then went directly to the add existing medication page and re-added the medication, it would not show up in the trip formulary; Fix Description: added inventoryService deleteMedication call in inventoryController.existingPost to un-delete medications properly if they had been deleted. Changed behavior such that the initial and current quantities set before the deletion are the same when the medication was re-added;

### DIFF
--- a/app/femr/business/services/system/MedicationService.java
+++ b/app/femr/business/services/system/MedicationService.java
@@ -88,7 +88,6 @@ public class MedicationService implements IMedicationService {
      * {@inheritDoc}
      */
     public ServiceResponse<MedicationItem> createMedication(String name, String form, List<MedicationItem.ActiveIngredient> activeIngredients) {
-
         ServiceResponse<MedicationItem> response = new ServiceResponse<>();
 
         try {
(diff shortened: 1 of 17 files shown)


app/femr/ui/controllers/admin/InventoryController.java
```diff
@@ -314,17 +314,35 @@ public Result existingPost(Integer tripId) {
...
 
                         return internalServerError();
                     }else{
 
-                        ServiceResponse<MedicationItem> setQuantityServiceResponse = inventoryService.setQuantityTotal(medicationItemServiceResponse.getResponseObject().getId(), tripId, 0);
+                        //If it exists in the formulary, delete or un-delete the medication from the formulary
+                        inventoryService.deleteInventoryMedication(medicationItemServiceResponse.getResponseObject().getId(), tripId);
+
+                        //Find the medication so we can find it's initial quantity, and current quantity
+                        ExpressionList<MedicationInventory> medicationInventoryExpressionList = QueryProvider.getMedicationInventoryQuery()
```
kevinzurek: Are you able to call the service layer instead of re-writing this?:

https://github.com/FEMR/femr/blob/master/app/femr/business/services/system/InventoryService.java#L117

app/femr/business/services/system/MedicationService.java
```diff
@@ -158,12 +157,13 @@ public MedicationService(IMedicationRepository medicationRepository,
             // There exist a matching medication in the database, so update that one rather then create new one
             if (matchingMedication != null) {
                 medicationRepository.deleteMedication(matchingMedication.getId(), false);
+                Ebean.save(matchingMedication);
```
kevinzurek: What is being saved here?

PR #631, opened 2018-03-08
After 90 days a user should be required to reset their password

fz1958:
Previously it was 60 days, now it is 90 days

https://teamfemr.atlassian.net/browse/FEMR-326


kevinzurek:
:+1: 

Merged manually:
4e69092b3349c5045b8c44916ed94e2d5f5d1725

PR #633, opened 2018-03-11
Joe Brown Change Request 1 - FEMR-325

AfroJoe759:
I edited about 4 files to ensure that the accepted password would have at least 8 characters.


kevinzurek:
:+1: 

Very good catch on the `CreateViewModel`

merged manually into master:
2b1a2d6cefaf58deb2f148707bc21fd99fbfc6dc

kevinzurek:
Created a new issue because the user javascript seems to be having errors when an admin creates a new user. The server side code from this PR catches the password length:

https://teamfemr.atlassian.net/projects/FEMR/issues/FEMR-340

PR #636, opened 2018-03-31
FEMR 330-331, Force logout after 30 minutes of inactivity

fz1958:


kevinzurek:
@fz1958 going to need some help sorting through which commits implement your change request - the pull request has over 100 commits, i think something went wrong during your merge!

PR #638, opened 2018-04-08
[FEMR-319] Upgrade Play from 2.5.18 to 2.6.12

kevinzurek:
*Now uses Akka HTTP server instead of Netty
*Configuration overrides added for managing idle and request timeouts
*h2 moved to external dependency, it was removed in Play 2.6
*Updated Scala from 0.13.11 to 0.13.15
*Changed Configuration to use typesafe config
*Keeping jodaForms instead of migrating to java.time
*Fixed deprecated crypto configuration
*Updated mysql connector
*Renamed sbt Build file and move it to root folder
*Got rid of deprecated value() function
*Upgraded Ebean and removed deprecated methods (findUnique)
*Use non-depcrecated version of WordUtils from apache commons text
*Fixed weird sbt error caused from capital letters in Build.sbt app name
*Removed mysterious js file in /app
*Changed asset routing to not use global state
*Updated gzip and digest
*Fixed forms to implement new validator interface

To be merged after some more testing. 2.6 enables reoccurring tasks to run in the application layer


kevinzurek:
Merged manually into master:

527b58d180a6c23fa9aeaac71677beff3fcb0f02

PR #642, opened 2018-04-13
FEMR 248

fz1958:
Edited the CSS to hide the data from the default, this way it could be replaced with a text box correctly displaying in the format dd/mm/yy.
All of this is done to make it easier for the User to use and understand.

https://teamfemr.atlassian.net/browse/FEMR-248


kevinzurek:
@fz1958 this is causing some issues when i try to apply filters in Research. See below:

image


I believe it's because you removed the name tag from the HTML elements

kevinzurek:
thanks for fixing that @fz1958 - merged manually :+1: 

3824d826129fdd4ab86f1a37ab5d8e84323233c9

PR #645, opened 2018-04-28
Fix error occuring with new inventory medication is added

anakhleh:
Bug occurred when a user tries to add a medication to the trip inventory that had never been added before (and thus cannot be readded because it has no soft-deletion state). Add new InventoryService function to check if a medication is related to the trip in question. Add logic in InventoryServiceController to use said medication and appropriately add new medication to trip inventory, or to just set the soft-delete state to 'undeleted'

https://teamfemr.atlassian.net/browse/FEMR-345

### DIFF
--- a/app/femr/business/services/core/IInventoryService.java
+++ b/app/femr/business/services/core/IInventoryService.java
@@ -89,7 +89,7 @@ public interface IInventoryService {
      * @param tripId id of the trip that will contain or contains the medication.
      * @return a medication item that contains quantity information.
      */
-    ServiceResponse<MedicationItem> createInventoryMedication(int medicationId, int tripId);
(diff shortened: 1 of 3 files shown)


kevinzurek:
Merged manually: removed the System.out calls and added a Logger to catch the stacktrace going forward. :+1: 

1cee5257f56d40a876d3bb2cffa9f4bf82ad79ba

There's a bigger issue at hand in regards to logging where the service layer is creating errors inside of the ServiceResponse<> object via `response.addError("", ex.getMessage());` and not doing any kind of logging to standard out. The controllers aren't smart enough to deal with ServiceResponse objects.

The result is masking the Java exception that originates in the service layer while the controller just passes an HTTP 500 response code

PR #647, opened 2018-04-30
FEMR 21 & FEMR 343

AfroJoe759:
FEMR 21 - Updating the conversions and limits for imperial and metric systems.

Update: Fixed the original problem with first pull request where only inches or centimeters converted correctly, not both if switched.

FEMR 343 - Fixing the error message bug in the edit user page when two roles are selected and when only one of the password fields is filled.

https://teamfemr.atlassian.net/browse/FEMR-343


kevinzurek:
Merged FEMR-343 manually:
9aac8c2d080c8560fff7dbf401ef1af1217c661f

Keeping FEMR-21 open so ebean doesn't get tightly coupled with the UI layer

PR #649, opened 2018-05-05
Fix fEMR-346 and fEMR-260

anakhleh:
fEMR-346: Provide brand names for medications where brand name was given as " ". Brand names that were whitespace would eventually cause a NullPointerException in MecationService.createMedication caused by from null being returned from MedicationRepository.createNewMedication (see https://github.com/FEMR/femr/blob/dddc749359d94786b923a08e46861500c46459cf/app/femr/data/daos/system/MedicationRepository.java#L55-L60), and then causing a NullPointerException.

fEMR-260: When all settings are unchecked, a null list of system settings is passed to ConfigureService.updateSystemSettings; Add logic in that method to account for that and set all settings to be inactive. Add logger message in ConfigureController for the (hopefully rare) case where config updates fail.

https://teamfemr.atlassian.net/browse/FEMR-346
https://teamfemr.atlassian.net/browse/FEMR-260

### DIFF
--- a/app/femr/business/services/core/IConfigureService.java
+++ b/app/femr/business/services/core/IConfigureService.java
@@ -35,7 +35,8 @@ public interface IConfigureService {
     /**
      * Update more than one system setting.
      *
-     * @param systemSettings a list of active system setting names. system settings that don't exist in this list are assumed to be inactive
+     * @param systemSettings a list of active system setting names. system settings that don't exist in this list are assumed to be inactive.
+     *                       null if all system settings are to be set to "off"
      * @return a list of system settings after the update.
      */
     ServiceResponse<List<? extends ISystemSetting>> updateSystemSettings(List<String> systemSettings);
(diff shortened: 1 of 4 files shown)


kevinzurek:
I merged FEMR-260 (error for not selecting any settings) manually after switching the logger to:

`Logger.error("ConfigureController-managePost()","Failed to update System Configuration Settings");`

to make it easier to trace back:
ed6bea0ebbfb1aaeff768afcb72d0bec607ee778

kevinzurek:
Still investigating FEMR-346 for a couple reasons:

1) The change won't impact existing systems, it will only update the concept_medications table for new databases. I'm thinking either an additional script or logic in the seeder can solve this. FEMR-346 wouldn't be solved for existing systems and database would be out of sync for the lookup table. 

2) While testing i noticed that an old database i have has 101 entries in the concept_medications table and a brand new one that i create only gets 96 entries. My thoughts are that the expected behavior here should be that the number of entries are equal since this is really just a lookup table. I don't have an answer for this yet.

kevinzurek:
Added some additional logic that makes sure existing databases only update the existing entry instead of adding new ones (the reason it went up from 96 to 101 entries in concept_medications):

https://github.com/FEMR/femr/compare/master...kevinzurek:FEMR346?expand=1

Still doing a bit more testing

kevinzurek:
Merged FEMR-346:

e8eeafeb7cf3c4f9b8e41072af5e05caa228c36e

PR #650, opened 2018-05-06
Add basic reverse ssh

anakhleh:
Add support for creating and maintaining a reverse ssh channel. Limited testing. May need more logging/exception handling

### DIFF
--- a/Build.sbt
+++ b/Build.sbt
@@ -23,7 +23,8 @@ val appDependencies = Seq(
   "com.google.code.gson" % "gson" % "2.3.1",
   "com.itextpdf" % "itextpdf" % "5.5.6",
   "com.itextpdf.tool" % "xmlworker" % "5.5.6",
-  "com.h2database" % "h2" % "1.4.193"
+  "com.h2database" % "h2" % "1.4.193",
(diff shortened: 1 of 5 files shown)


kevinzurek:
This is awesome. After testing for a while i added a commit:

https://github.com/kevinzurek/femr/commit/e03d179f87e83e1a42e7c1c02cf77ad8c3e579ac

Not much changes other than re-using the logic to build and maintain the SSH session. Added some more verbose informational logging and an additional configuration item. 

`remoteSshPort` = the SSH port on the central server that is listening for a connection from a client
`remoteSshPortForward` = the port on the central server that will forward the SSH session back to a client (reverse tunnel)
`localSshPort` = the port on the client that will accept the reverse tunnel

I set up port 5122 on fEMR central for sshd to listen on. The only thing i'm still interested in testing is 2 devices both using port 5122 to create the initial session, but forwarding different ports back to themselves at the same time

kevinzurek:
"basic" reverse ssh

kevinzurek:
Merged manually:

10ea6fe02942d0fff75edfbea2648d840d2d7381

:100:

PR #652, opened 2018-05-12
Fix recognition of when rssh has failed

anakhleh:
When rssh fails, the build function returns null. Properly check for null, and rebuild the rssh if the ssh session is not null but is also not connected


kevinzurek:
Merged manually:
3aa66dc58c2a55dbf5b29e2871d285c32d8452c4

PR #653, opened 2018-05-14
Fix jsch spawning lots of sshd processes on fEMR central

anakhleh:
Occurs when an ssh connection is made, but reverse port forwarding fails.


kevinzurek:
Merged both manually:
89d34df45f0ef0ff0baf086fb6c9096713fa2ddb
a9f92e0a8419927ab8f165ffe8cec0e1459725bc
:+1:

PR #657, opened 2018-12-13
Possible Fix to fEMR-353 and other encounter pdf improvements

anakhleh:
What has been done:

1) properly displayed metric and imperial units
2) Chief Complaint section under encounters
3) Blood pressure only displayed in blood pressure section. Heart rate displayed in heart rate section
4) Changed Chief Complaints section to read 'HPI & Chief Complaint'
5) Change Problems, prescription, vitals sections to show N/A when no data
6) Change Vitals to display the time taken, in the case where it it taken multiple times a day.

What has not been done:
1) full testing with full patient data like in the field.

What I'm not sure about:
1) A User group reported that the HPI section did not display in some cases. Having checked the medical HPI tab, it looks like all the data that should be there is there. Running the HPI tab with only the narrative section, does not cause it not to display. Perhaps there is some particular program state that causes this that I have not encountered.
2) I have tested it only a few times when in Imperial unit mode. So far it has worked as intended, but has not been tested as much as when in metric mode

### DIFF
--- a/app/femr/ui/controllers/PDFController.java
+++ b/app/femr/ui/controllers/PDFController.java
@@ -334,17 +334,17 @@ public class PDFController extends Controller {
         table.addCell(getVitalMapCell("Blood Pressure: ", "bloodPressure", vitalMap));
         table.addCell(getVitalMapCell("Temperature (°F):", "temperature", vitalMap));
         table.addCell(getVitalMapCell("Temperature (°C):", "temperatureCelsius", vitalMap));
-        table.completeRow();
+//        table.completeRow();
 
         table.addCell(getVitalMapCell("Glucose:", "glucose", vitalMap));
         table.addCell(getVitalMapCell("Height (Imperial):", "heightImperial", vitalMap));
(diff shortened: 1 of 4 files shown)


kevinzurek:
Hey @anakhleh !

Looks good, i think it's mergeable as-is - i like outputting metric & imperial strings instead of having to check the current system configuration :+1: 

Also, I wasn't able to reproduce # 1 about the HPI section either. At first, i thought i would be able to by inputting a combination of things while the patient has an open encounter:
1. multiple chief complaints
2. toggling the system configuration option for allowing multiple chief complaints
3. toggling the system configuration option for consolidating the HPI tab and filling out the narrative section

All combinations of the above produced expected behavior (comprehensive output on the pdf)

Only feedback i have is minor:
1. probably could do without seconds & milliseconds on the vitals timestamp
2. the changes to `application.conf` and `application.example.conf` were committed a bit ago already: https://github.com/FEMR/femr/commit/5a99f47e9d795ed108c7e7bbc0b07b715fc02e0d
3. Damn, the PDFController has clean and readable code from 4 years ago. Nice @kdunlap :)

PR #658, opened 2018-12-21
femr-353 Proposed Fix revised - no seconds in timestamps

anakhleh:
Pretty much the old fix, but now timestamps display as YYYY-MM-DD HH:mm instead of YYYY-MM-DD HH:mm:ss as in the previous fix. PDF's of the updated fix are in the github_integration slack channel


kdunlap:
Merged manually: 

c3bb39c447c7824b7ebec38df3e87c8554c6cb65

PR #661, opened 2019-07-18
Triage validation improvements

kdunlap:
### Updates Completed
- tried to simplify validation code, used css classes for errors rather than styles in js
- made errors highlight both the field and the label
- made errors disappear if fixed on the second submit button click
- added vital validation with ranges **[FEMR-360]**
	- on both Triage and Medical Tabs
	- added error messages for vital validation to show the valid ranges
- updated Age Validation **[FEMR-280][FEMR-117]**
	- cannot enter age in the future
	- entered age and age group must match if both are present 
	- added message when age and age group do not match
	- users can enter only inches **[FEMR-321]**
	- prevented birth date from being a date in the future
- enchanced height input on front end **[FEMR-21]**
	- users can enter inches > 12 (or cm > 100) and get the appropriate conversion
- selecting a gender is now required

Also rearraged ci files, moved existing to `concourse` folder and added `aws` folder
- Added `demo.conf` for demo deployment environment vars
- added `buildspec.yml` for CodeBuild
- started `README.md` for CodePipeline, CodeBuild, ElasticBeanstalk setup of demo

https://teamfemr.atlassian.net/browse/FEMR-280
https://teamfemr.atlassian.net/browse/FEMR-117
https://teamfemr.atlassian.net/browse/FEMR-321
https://teamfemr.atlassian.net/browse/FEMR-21
https://teamfemr.atlassian.net/browse/FEMR-360


anakhleh:
As of 2:10pm July 20, 2019:

On reviewing the JS code itself, it looks organized and clean to me. Could have a few extract function refactors, but that is more a point of personal preference.

As for the fix of FEMR-280, Validation could be more extensive.
-  First, text and negative numbers can be entered into the Years and Months inputs, and if an age category is chosen, then the form can submit given that all other requirements are fulfilled. In the patient page with title 'General Info' at route /history/patients/{patient id} the age value is provided as empty (the bit about the general info page is expected behaviour given how the triage form is filled out, but being able to submit the triage form under those circumstances should not be).
-  Second, it is possible to select a birth date in the future through the Birth Date input (ex. you can enter a 2020 date), and the form will submit so long as a age classification checkbox is checked. In the General info page, the patient's age will be recorded as a negative number of months.

As for required gender, it seems to be working as intended.

kdunlap:
Thanks for taking a look at this @anakhleh. It is much appreciated!

Good find on the negatives. I get your point with the extract function refactors, it was giving me an itch as well, but I didn't want to introduce too much change without front end tests. I haven't written jQuery in about 3 years haha. Fingers crossed we can refactor the UI one day to use a more modern reactive framework.

I also added validation on the vital measurements on both the Triage and Medical pages. I'll add some details to the notes on the PR of what changed to make testing there easier. I'll also update the things you found, and maybe you can take another quick look next week.

Thanks again!

public/js/triage/triage.js
```diff
@@ -384,6 +384,7 @@ var multipleChiefComplaintFeature = {
 var triageFields = {
 
     patientInformation: {
+        firstName: $('#firstName'),
```
kdunlap: @kdunlap - Note to self to remove this extra line, not sure how it got there

anakhleh:
As of 6:28 on July 22, 2019:

I quickly tested the triage code and I think it's substantially more robust than before. That being said, there were a cases I had questions about:

- It is still possible to enter 0 for heart rate, respiration, and oxygen saturation. I assume that this makes sense in the case that perhaps a severely ill person or even recently deceased person is brought to a clinic where the real-world values for these could be 0.
- While you can't submit the form when this happens (the validation is robust in this case), it is kind of amusing that providing a meters height (x given x > 0) and a negative centimeters height ((-1)*(100*x-1) given x > 0 )leads to the BMI calculation being absolutely massive.
- It is possible to submit a patient who is 1 cm tall, but is 500kg heavy. That being said, I'm not convinced this user error occurs often, and I think fixing this would be very involved in code. You would need some kind of mapping from possible heights to weights and vice versa, but then what happens when you get an outlier such as a very malnourished child and then the validation won't accept the valid but non-ordinary data?. It is possible then that this should actually not be fixed.

Overall, I'd still say the validation is pretty good, and that these are only minor cases I found that I wanted to bring up just in case.

kdunlap:
Manually merged

PR #669, opened 2021-10-14
Super-femr 

JayantDevkar:
cloning repo from capstone to femr repo


kevinzurek:
Ran into this while trying to use `super-femr` https://github.com/FEMR/femr/issues/686

PR #673, opened 2021-12-01
Trauma button added to nav bar

17rey17:
Added Trauma button to the nav bar.
Screen Shot 2021-11-30 at 7 26 23 PM


Baagi:
I think this needs to be rebased on super-femr branch instead of the master branch?

PR #680, opened 2022-03-08
UI updates

sky-kurth:
Basic UI style changes, minor refactoring of input fields

### DIFF
--- a/app/femr/ui/views/trauma/index.scala.html
+++ b/app/femr/ui/views/trauma/index.scala.html
@@ -33,7 +33,13 @@
                 <div id="leftGeneralInfo" >
 
                     <h4>Patient Age</h4>
-                    <input type="checkbox" id="1" name="1" value="Age1">
+                    <div id="classificationRadioWrap">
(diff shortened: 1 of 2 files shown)


Baagi:
@sky-kurth I think the changes in [public/css/femr.css](https://github.com/FEMR/femr/pull/680/files/544f351477bcdad16b1188d017ecb59171155ff7#diff-2163e550d12f4aa3b97ba01d76d4e6438fa6785609773e6b5cfdfbac48917492) are nice but the style changes are affecting the whole application. This looks like a major UI change and it would be better to confirm if the femr team wants to have these colors. 

I personally would just use the current styles so that it doesn't affect other pages.  

Screen Shot 2022-03-10 at 9 17 25 AM


The red outline looks great with the theme color but it kind of feels like the red outline is showing up because the user input is wrong.
Screen Shot 2022-03-10 at 9 23 47 AM

Baagi:
Looks good to me! Going to merge this to `tremr-branch`

PR #681, opened 2022-03-08
Tremr: Disable pregnancy weeks if not female

Baagi:
## Description
If the female checkbox is selected, then enable the pregnancy weeks input, otherwise, disables the input.

Also added now only one checkbox can be checked for the age and gender inputs.


sky-kurth (commented):
The checkbox code should probably be built using the existing structures found on the triage intake page, which looks like:

`<div id="classificationRadioWrap">
                    @for((key, valyew) <- viewModel.getPossibleAgeClassifications) {
                        <label><input type="radio" name="ageClassification" value="@key" /><span>@valyew</span></label>
                    }
                    </div>`

It can be found in ui/view/triage/index.html for reference

Baagi:
Sounds good. I reused the inputAge() field with slight modifications. 

Screen Shot 2022-03-13 at 5 31 26 PM

PR #683, opened 2022-03-14
Rey's branch

17rey17:


app/femr/ui/views/trauma/index.scala.html
```diff
@@ -341,6 +341,8 @@ <h4>Vitals Chart</h4>
                         <input type="checkbox" id="drawCheckbox" value="brush" name="drawCheckbox" onclick="onlyOne(this)" checked>
                         <label for="eraser">Eraser</label>
                         <input type="checkbox" id="drawCheckbox" value="eraser" name="drawCheckbox" onclick="onlyOne(this)">
+                        <button id="save" onclick="downloadURI(this)">Save Image</button>
```
Baagi, line 340:
```suggestion
                        <button id="save">Save Image</button>
```

Since you already added event listener in trauma.js, we don't need to add it here. https://github.com/FEMR/femr/blob/3383c9de476f72ca18083b35892dda21791b6174/public/js/trauma/trauma.js#L24-L31

Baagi: Otherwise it tries to download 2 times like this:

Screen Shot 2022-03-13 at 7 36 25 PM

PR #698, opened 2022-05-05
Tremr: Added Undo button for lines/drawings

Baagi:
https://user-images.githubusercontent.com/23184246/166971974-382e2f1a-0418-4488-8a8a-d8f99ad5ac1f.mov


sky-kurth:
Nice! Might be better to rename some variables for clarity, but this looks very clean :)

PR #701, opened 2022-05-19
trauma vitals stored via sessionstorage

sky-kurth:
Vitals can be logged as sets to sessionnstorage before patient is saved to DB

### DIFF
--- a/app/femr/ui/models/trauma/IndexViewModelGet.java
+++ b/app/femr/ui/models/trauma/IndexViewModelGet.java
@@ -53,6 +53,10 @@ public class IndexViewModelGet {
         return vitalNames;
     }
 
+    public List<VitalItem> getVitalNamesForTable() {
+        return vitalNames.subList(0, 11);
+    }
+
     public void setVitalNames(List<VitalItem> vitalNames) {
         this.vitalNames = vitalNames;
     }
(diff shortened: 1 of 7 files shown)


app/femr/ui/views/trauma/index.scala.html
```diff
@@ -138,7 +138,7 @@ <h4>Vitals</h4>
                                     <input type="number" step="any" min="0" class="fInput" id="@viewModel.getVitalNames.get(3).getName" name="@viewModel.getVitalNames.get(3).getName" placeholder="%"/>
                                 </div>
                                 <div class="vitalWrap">
-                                    <label for="heightFeet">Height</label>
+                                    <label for="height">Height</label>
```
Baagi, line 141: Why was this changed? Looks like `heightFeet` is used in many places in the code when I do `git grep heightFeet`.

app/femr/ui/models/trauma/IndexViewModelGet.java
```diff
@@ -53,6 +53,10 @@ public List<VitalItem> getVitalNames() {
         return vitalNames;
     }
 
+    public List<VitalItem> getVitalNamesForTable() {
+        return vitalNames.subList(0, 11);
+    }
+
```
Baagi, line 59:
```suggestion

```

PR #713, opened 2024-02-27
Refactored the translation flow and fixed unavailable port errors (and minor speedtest changes)

mhayes2772:


DJump13:
Moving the port selection away from a hardcoded url into another file specifically for the translation server was a great idea, this looks like a perfect fix for the port bug.

translator/server.py
```diff
@@ -12,11 +12,14 @@
 from transformers import MarianMTModel, MarianTokenizer
 from typing import Sequence
 from libargos import install_packages
+import socket
 
 PORTS = [8000, 5000]
```
noah1013, line 17: Maybe add more ports. If we have 8-10 or so ports, theres no scenario except for testing where all ports will be used up.

PR #714, opened 2024-03-07
Added 'English in the Middle' functionality to server.py

noah1013:
Allows translations between languages to be done with Argos if we have two packages that both translate to English.

### DIFF
--- a/translator/server.py
+++ b/translator/server.py
@@ -42,21 +42,23 @@ class WebRequestHandler(BaseHTTPRequestHandler):
         from_code = self.query_data['from']
         to_code = self.query_data['to']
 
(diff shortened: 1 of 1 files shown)


mhayes2772 (commented):
We might want to make sure that the English -> to_code path exists, in case somehow we end up with a language that has the From -> English path but not the other direction

DJump13:
Good point Michael, for both argos and marian too, but otherwise looks clean

noah1013:
Gotcha, I was thinking that because we have all Argos packages downloaded it doesn't matter which order of the packages we have, but if new packages are made for Argos and we download them, I see how order does matter.

PR #715, opened 2024-03-07
Button Changes

DJump13:
Added toggle functionality to show original button, included libraries and base code to make JSON parsing for other text fields easier

### DIFF
--- a/translator/server.py
+++ b/translator/server.py
@@ -56,7 +56,7 @@ class WebRequestHandler(BaseHTTPRequestHandler):
                 translatedText = marian.translate([text])
                 return translatedText[0]
             else:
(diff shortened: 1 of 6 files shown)


translator/server.py
```diff
@@ -56,7 +56,7 @@ def translate_data(self):
                 translatedText = marian.translate([text])
                 return translatedText[0]
             else:
-                return "Translation Unavailable"
+                return from_code + to_code + "Translation Unavailable"
```
noah1013, line 59: I like the idea for this change. I think if we do something like this, it should include the actual languages instead of the ISO language codes so the user knows the translation didn't work and with what languages.

PR #716, opened 2024-04-09
Added Language Codes to Mock Java Files 

noah1013:
These language codes were added in their respective interface but not in the concrete classes that implemented them. Adding these codes reenables the testing suite for fEMR.


DJump13:
I think all that's needed is to add a private String languageCode at the top of each class so the mocks can handle those tests

DJump13:
I think this is all set

PR #718, opened 2024-04-16
Added RTL css for Right to Left Languages

Har150n:
When the Show Original button is clicked, the Right to Leftness is maintained. Found a bug where if Arabic or Hebrew is inputted, the web app breaks because it can't be added into the URL parameter of the translate


SeanHershey (changes requested):
For me the fields are not filling in Arabic but the direction is correct and toggles properly.

Har150n:
> For me the fields are not filling in Arabic but the direction is correct and toggles properly.

think that's a delimiter issue, the translation is lumping fields together

SeanHershey (commented):
Then it looks good. Just in time for delim issue because I found the @ symbol to be way better the $ so I'll change that soon.

PR #725, opened 2024-05-15
Created LICENSE checker script

noah1013:
Added the license_check.yml script that prevents new pushes without the LICENSE file existing in the main femr directory. Also added the LICENSE script to the branch.


.github/workflows/license_check.yml
```diff
@@ -0,0 +1,20 @@
+name: Check for LICENSE File in femr Directory
+
+on:
+  push:
+    branches:
+      - temr-dev-noah
```
mhayes2772: I think maybe you can just remove the "branches:" portion, and it will do it for any push regardless of branch? Not 100%

mhayes2772 (approved):
Nice work

PR #767, opened 2025-02-19
Added waits to ensure loading & added E2E tests for request to join + admin approval

JLpro-cd:


integration_test/tests/test_basic_functionality.py
```diff
@@ -156,4 +162,140 @@ def test_search(driver):
...
+
+
+
+
+
+
+
+
+
+
```
realkaranvir, line 301: Everything looks good to me but let’s remove these extra lines

realkaranvir (changes requested):
Looks good to me. The selenium tests are probably failing right now because they aren’t using this change until you merge into master. Remove the white space and it’s good

integration_test/tests/test_basic_functionality.py
realkaranvir: Looking at the failed workflow run for selenium tests it looks different than previously… Can we check that it’s not failing for a different reason this time?

PR #776, opened 2025-03-06
FHIR support for body temp, blood pressure, weight

bholt13:


bholt13:
some sort of issue with prescriptions being null. builds fine locally... we'll discuss in person 3/6/2025

PR #786, opened 2025-04-10
Patient month and years (age) is now properly calculated from a given birthdate in triage

JLpro-cd:
let's go


JLpro-cd:
Fixing

JLpro-cd:
Should be good now

JLpro-cd:
Yes. It immediately updates the age numbers based on date, same as if they manually inputted the numbers themselves.

PR #791, opened 2025-04-29
fix(research): allow primary dataset to be selected

StarDylan:
The "id" column is supposed to be called `data-dname1` as referenced by filter-menu.js.


bholt13 (approved):
looks good

PR #792, opened 2025-04-29
Kit versioning through GHA workflow

realkaranvir:
Versioning done through GHA workflow. Need to update all references to kit version in the app still. Please review thoroughly.


thekevinWHYuan (approved):
LGTM!

bholt13 (approved):
Approved for testing

Build.sbt
```diff
@@ -6,7 +6,7 @@ import com.typesafe.sbt.web.Import._
 
 
 val appName = "femr"
-val appVersion = "3.0.0"
+val appVersion = "3.1.1"
```
StarDylan, line 9: Maybe remove this since we're getting it from ENV?

app/femr/util/startup/DatabaseSeeder.java
```diff
@@ -151,7 +153,11 @@ private void seedKitStatus() {
         if (kitStatuses != null && !containKitStatus(kitStatuses, "Version")) {
             kitStatus = new KitStatus();
             kitStatus.setName("Version");
-            kitStatus.setValue("1.0.0");
+            String version = System.getenv("APP_VERSION");
+            if (version == null || version.isEmpty()) {
+                version = "Test Version";
```
StarDylan, line 158: You could maybe try to add the commit hash instead of just test version

StarDylan (commented):
Few comments, but the core part of it looks good to me

app/femr/util/startup/DatabaseSeeder.java
realkaranvir: This case would only happen during our push where it tests the build though.

realkaranvir: This case would only happen during our push where it tests the build though.

Build.sbt
realkaranvir: Honestly I don't even know what this part does. This isn't where our femr version number is set in the app it just sets the version number for the SBT build ??

PR #798, opened 2025-05-28
DB Dump

thekevinWHYuan:
oops should've probably made a pr for this
- Implemented button to create a backup/db dump
- Made it so it would only work if SQL CLI tool is installed on the user machine


realkaranvir (approved):
lgtm

PR #803, opened 2025-06-06
Removed Unknown Version from developer view

bholt13:


app/femr/business/services/system/FhirExportService.java
thekevinWHYuan, line 1: What's the change in FhirExportService for?

realkaranvir: i'm assuming it's for the welc assignment?

realkaranvir (approved):
code looks good to me.

PR #804, opened 2025-06-08
Cleared list which was causing errors displaying trip name

JLpro-cd:
For some reason this fixes it... We make the list before and it should be empty but there was some undefined behavior with it sometimes being empty and sometimes not... This works now.


bholt13 (approved):
looks good ... easier to test now ... nice

PR #806, opened 2025-11-12
Add basic DailyReport class

KrakenMInitials:
Initial class to act as an interface between obtaining data for MDS and generating actual Daily Report file


KrakenMInitials:
DailyReport class outdated, unused, and archived.

PR #808, opened 2026-01-22
Perf/optimize research export

KrakenMInitials:
Main Issue: One of the fEMR deployments crashed during research export. The issue is hypothesized to be due to limited compute resources and lack of RAM in the machines with fEMR.

Solution: Identified duplicate data representations and streamlined export file creation removing any redundancies in data transformations.

Observations: 
- significantly cut down time for export production from ~30 seconds to ~8 seconds.
- inconclusive whether RAM improvements were made 

Notes
- Due to deprecated openjdk:8-jre-alpine base docker image, found workaround with eclipse-temurin:8-jre-alpine-3.23. 
 - Added minor UX improvements to Research Export page but may be overridden with UI/UX reworks
 - New research export code makes use of OS configured temp directory to build file and set file as response.
 - Replacement with eclipse-temurin image causes file permission issues when old code is built into a docker container and research export is ran. PR essentially also contains fixes to research export beyond docker image replacement.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -67,7 +67,7 @@ RUN unzip femr-*.zip && rm femr-*.zip
 FROM eclipse-temurin:8-jre-alpine-3.23
 
 RUN apk update
-#RUN apk add --no-cache bash python3 py3-pip gcc python3-dev musl-dev linux-headers mariadb-client mariadb-connector-c mariadb-connector-c-dev
+#RUN apk add --no-cache bash python3 py3-pip gcc python3-dev musl       -dev linux-headers mariadb-client mariadb-connector-c mariadb-connector-c-dev
 #RUN pip3 install psutil
 RUN apk add --no-cache bash python3 py3-pip py3-psutil gcc python3-dev musl-dev linux-headers mariadb-client mariadb-connector-c mariadb-connector-c-dev
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -64,13 +64,12 @@ RUN --mount=type=cache,target=/root/.ivy2 \
 WORKDIR $PROJECT_HOME/app/target/universal
 RUN unzip femr-*.zip && rm femr-*.zip
 
-#FROM openjdk:8-jre-alpine
 FROM eclipse-temurin:8-jre-alpine-3.23
+
 RUN apk update
-#RUN apk add --no-cache bash python3 py3-pip py3-psutil gcc python3-dev musl-dev linux-headers mariadb-client mariadb-connector-c mariadb-connector-c-dev
+#RUN apk add --no-cache bash python3 py3-pip gcc python3-dev musl-dev linux-headers mariadb-client mariadb-connector-c mariadb-connector-c-dev
 #RUN pip3 install psutil
 RUN apk add --no-cache bash python3 py3-pip py3-psutil gcc python3-dev musl-dev linux-headers mariadb-client mariadb-connector-c mariadb-connector-c-dev
-
 
 #database variables
 ARG APP_VERSION
@@ -88,4 +87,4 @@ EXPOSE 9000
 
 # run fEMR using env variables
 #ENTRYPOINT url=$DB_URL usr=$DB_USER pass=$DB_PASS sbt ~run
-ENTRYPOINT ["/bin/bash", "-c", "/opt/bin/femr/bin/femr"]
+ENTRYPOINT ["/bin/bash", "-c", "rm -f /opt/bin/femr/RUNNING_PID; exec /opt/bin/femr/bin/femr -Dpidfile.path=/dev/null"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -64,11 +64,13 @@ RUN --mount=type=cache,target=/root/.ivy2 \
 WORKDIR $PROJECT_HOME/app/target/universal
 RUN unzip femr-*.zip && rm femr-*.zip
 
-FROM openjdk:8-jre-alpine
-
+#FROM openjdk:8-jre-alpine
+FROM eclipse-temurin:8-jre-alpine-3.23
 RUN apk update
-RUN apk add --no-cache bash python3 py3-pip gcc python3-dev musl-dev linux-headers mariadb-client mariadb-connector-c mariadb-connector-c-dev
-RUN pip3 install psutil
+#RUN apk add --no-cache bash python3 py3-pip py3-psutil gcc python3-dev musl-dev linux-headers mariadb-client mariadb-connector-c mariadb-connector-c-dev
+#RUN pip3 install psutil
+RUN apk add --no-cache bash python3 py3-pip py3-psutil gcc python3-dev musl-dev linux-headers mariadb-client mariadb-connector-c mariadb-connector-c-dev
+
 
 #database variables
 ARG APP_VERSION
--- a/app/femr/business/helpers/LogicDoer.java
+++ b/app/femr/business/helpers/LogicDoer.java
@@ -93,9 +93,9 @@ public class LogicDoer {
                 path += File.separator;
             return path;
         } catch (Exception ex) {
-            //If config doesn't exist, default to "photos"
-            path = "../Upload/Csv";
-            return path;
+            //If config doesn't exist, default to "../Upload/CSV"
+            return ".." + File.separator + "Upload" + File.separator + "CSV" + File.separator;
+
         }
 
     }
--- a/app/femr/business/services/system/ResearchService.java
+++ b/app/femr/business/services/system/ResearchService.java
@@ -50,6 +50,7 @@ import java.io.IOException;
 import java.text.ParseException;
 import java.text.SimpleDateFormat;
 import java.util.*;
+import play.Logger;
 
 public class ResearchService implements IResearchService {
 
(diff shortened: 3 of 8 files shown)


Dockerfile
```diff
@@ -64,11 +64,13 @@ RUN --mount=type=cache,target=/root/.ivy2 \
 WORKDIR $PROJECT_HOME/app/target/universal
 RUN unzip femr-*.zip && rm femr-*.zip
 
-FROM openjdk:8-jre-alpine
-
+#FROM openjdk:8-jre-alpine
```
lab596: No longer needed since actual fix has been merged now

app/femr/business/helpers/LogicDoer.java
```diff
@@ -93,9 +93,9 @@ public static String getCsvFilePath() {
                 path += File.separator;
             return path;
         } catch (Exception ex) {
-            //If config doesn't exist, default to "photos"
-            path = "../Upload/Csv";
-            return path;
+            //If config doesn't exist, default to "../Upload/CSV"
+            return ".." + File.separator + "Upload" + File.separator + "CSV" + File.separator;
```
lab596, line 97: why are we prefering this over the exact filepath? All other exceptions use exact file path, either match pattern or explain why this is needed

app/femr/ui/controllers/ResearchController.java
```diff
@@ -83,10 +83,11 @@ public Result indexGet() {
         filterViewModel.setMissionTrips(missionItemServiceResponse.getResponseObject());
 
         // Set Default Start (30 Days Ago) and End Date (Today)
+        // temporarily "-365 * 10" modified for development to default to last ten years
```
lab596, line 86: why? how temperary add a TODO maybe

app/femr/ui/controllers/ResearchController.java
```diff
@@ -130,6 +131,9 @@ public Result exportPost() {
         // This does weird stuff and isn't reliable.
         //ServiceResponse<File> exportServiceResponse = researchService.retrieveCsvExportFile(filterItem);
         ServiceResponse<File> exportServiceResponse = researchService.exportPatientsByTrip(filterItem.getMissionTripId());
```
lab596, line 133: nit: not sure about how researchService is written but this should throw an exception on an error not result in null. Here we should catch the error and wrap if needed or allow it to bubble up. Usually catching null and throwing and error based on that is not good practice. Also if you want to make sure fields are not null in java u usually dont want to use a null check and instead use @NotNull - view this: https://www.baeldung.com/java-notnull-method-parameter

public/js/research/filter-menu.js
```diff
@@ -157,10 +157,60 @@ var filterMenuModule = (function () {
...
     };
 
-    var exportData = function () {
+    var exportData = async function (e) {
+      if (e) {
+        e.preventDefault();
+        e.stopPropagation();
+      }
+
+      const btn = document.getElementById("export-button");
```
lab596, line 166: this seems a bit fragile why are we fetching the button by its element instead of passing it in as a param

app/femr/business/services/system/ResearchService.java
```diff
@@ -43,13 +43,17 @@
 import femr.util.stringhelpers.CSVWriterGson;
 import femr.util.stringhelpers.GsonFlattener;
 import femr.util.stringhelpers.StringUtils;
+import org.apache.commons.csv.CSVFormat;
+import org.apache.commons.csv.CSVPrinter;
 
-import java.io.File;
-import java.io.FileNotFoundException;
-import java.io.IOException;
+import java.io.*;
```
lab596: why all here instead of explicit imports, explicit is almost always better

app/femr/business/services/system/ResearchService.java
```diff
@@ -198,14 +202,13 @@ private ResearchExportItem createResearchExportItem(IResearchEncounter encounter
         ResearchExportItem exportitem = new ResearchExportItem();
 
         IPatient patient = encounter.getPatient();
-
-        // Patient Id
-        exportitem.setPatientId(patientId);
```
lab596, line 203: any reason this same exact code is moved after the age?

app/femr/business/services/system/ResearchService.java
```diff
@@ -342,72 +345,361 @@ else if (filters.getPrimaryDataset().equals("prescribedMeds") ||
...
+                        item.getVitalMap().get("alcohol"),     // alcohol
+                        item.getTabFieldMap().get("assessment"), // assessment
+                        item.getVitalMap().get("bloodPressureDiastolic"), // bloodPressureDiastolic
+                        item.getVitalMap().get("bloodPressureSystolic"),  // bloodPressureSystolic
+                        item.getChiefComplaints().isEmpty()
+                                ? null
+                                : item.getChiefComplaints().get(0),       // chiefComplaints1
+                        item.getTabFieldMap().get("currentMedication"),   // currentMedication
+                        item.getDayOfVisit(),                             // dayOfVisit
+                        item.getVitalMap().get("diabetic"),               // diabetic
```
BrendannnD, line 531: Calling .get() on string identifiers is not ideal because the schema could change.

app/femr/business/services/system/ResearchService.java
```diff
@@ -342,72 +345,361 @@ else if (filters.getPrimaryDataset().equals("prescribedMeds") ||
...
+            }
+
+        try {
+            printer = new CSVPrinter(writer, fileformat);
+        } catch (Exception e) {
+            Logger.error("ResearchService:exportPatientsByTrip CSVPrinter could not be instantiated:");
+            Logger.info("user.dir=" + System.getProperty("user.dir"));
+            Logger.info("exportFile=" + exportFile.getAbsolutePath());
+            e.printStackTrace();
+            return null;
```
lab596, line 457: this error is being swallowed here if we return null

app/femr/business/services/system/ResearchService.java
```diff
@@ -342,72 +345,361 @@ else if (filters.getPrimaryDataset().equals("prescribedMeds") ||
...
+        }
+        try {
+            printer.close();
+        } catch (Exception e) {
+            Logger.error("ResearchService:exportPatientsByTrip exception occurred attempting to close printer.");
+            e.printStackTrace();
         }
 
-        File eFile = createCsvFile(researchExportItemsForCSVExport);
+        response.setResponseObject(exportFile);
```
BrendannnD, line 603: No automated testing for the new CSV processing. How did you determine the time was faster as a result of optimization and not just caching?

app/femr/business/services/system/ResearchService.java
```diff
@@ -342,72 +345,361 @@ else if (filters.getPrimaryDataset().equals("prescribedMeds") ||
...
+        } catch (Exception e)
+        {
+            e.printStackTrace();
+            return response;
+        }
 
         return response;
     }
 
+    // @Override
```
j-cavender: Old code commented out to get rid of

Dockerfile
KrakenMInitials: Cherry picked fix commit hash and pushed changes. Should be fixed

app/femr/business/helpers/LogicDoer.java
KrakenMInitials: Different file systems use different file separators, so I believe this was the less OS-dependent way to do things just out of practice. If desired, can revert back to just using "/"

app/femr/business/services/system/ResearchService.java
KrakenMInitials: Resolved 😭

app/femr/business/services/system/ResearchService.java
KrakenMInitials: Time measurements were obtained using print statements and timers. Valid concern and automated testing to be added and verified next sprint or sometime.

KrakenMInitials: Backlog issue opened

app/femr/ui/controllers/ResearchController.java
KrakenMInitials: Code changes just changes the default option of the date filter on the Reserach section to show the last 10 years, should have no significant impact on feature but better qol perhaps by showing stats for the entire trip (inc. the trip isn't > 10 years.

public/js/research/filter-menu.js
KrakenMInitials: Valid concern but this might be out of scope. The button is passed from the frontend as "export-button" which is also referenced in multiple places.
Comment not resolved

app/femr/ui/controllers/ResearchController.java
KrakenMInitials: researchService returns null on internal service error. The research controller catches and handles the error by creating an Error response on the page. 
This is lazy checking, but on the positive side, the old code did not check for null or any type safeties at all and ran on hopes and prayers. Tried to keep scope to fixing performance and memory issues and kept most code design the same.

app/femr/business/services/system/ResearchService.java
KrakenMInitials: The research controller handles and raises and error on null response. Open to ears on better practices in error handling

PR #811, opened 2026-02-24
Reconfigure docker CI/CD workflows

KrakenMInitials:


KrakenMInitials:
Comment: not safe, all branches and pr pushed to github that triggers github action will push to dockerhub

KrakenMInitials:
Created an additional workflow "publish-docker-image.yml" that follows same logic but includes pushing image to Dockerhub triggered on pushes to master branch. Pushes to main will now automatically modify image on docker hub. Merges into main will require extra care.

pulish-docker-image.yml and docker-image.yml are mutually exclusive.

PR #821, opened 2026-05-15
Add speedtest scripts to docker image

KrakenMInitials:
While building the docker image, the Dockerfile doesn't include ./speedtest scripts into the final app image. Not sure how the current builds were handling speedtest functions if they were working. 

Discovered while working on FEMR/windows-installer. The ./speedtest folder from FEMR/femr is required to be manually copied into the FEMR/windows-installer to generate installers and ./speedtest is part of femr app logic. It would be more organized to directly have the speedtest scripts be a part of the image.

Currently, worked around but would like a merge asap to keep windows-installer organized and update femr image on docker hub through the Github Actions.


BrendannnD (approved):
Looks great

PR #823, opened 2026-05-27
Quick workflow patch

KrakenMInitials:
Removed arm64 build temporarily, will require future attention or new base docker image replacement in Dockerfile. 

eclipse-temurin:8-jre-alpine-3.23 does not support arm64 architecture.


KrakenMInitials (commented):
Looks good

BrendannnD (approved):

